PR #62, opened 2015-01-01
fix debounce opt on watch task

leblowl:
boot was throwing an error attempting to look up type long in boot.cli/assert-atom & clojure ints are longs anyways. ps what's with the ghost new line change below @ line 251?


micha:
Awesome, thanks! I think my vim setup sometimes adds whitespace like that on empty lines...

PR #119, opened 2015-02-03
warn when the version of required dependencies change

schmir:
(I'm still opening pull requests as I'm not quite sure if there's a better way to check for the problem here with aether.)

I've ran into an issue today. My project has a dependency on
ring/ring-devel "1.3.1", which itself depends on
org.clojure/tools.namespace "0.2.4"

When I start a development repl, I add the org.clojure/tools.namespace
"0.2.9" dependency. boot didn't warn me about that and silently used the
old version of tools.namespace (which of course had a bug).

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -175,10 +175,34 @@
   (#'clojure.core/load-data-readers)
   (set! *data-readers* (.getRawRoot #'*data-readers*)))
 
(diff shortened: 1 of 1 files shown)


micha:
Thanks! I think this problem might be more complicated, though, when you consider transitive dependencies...

schmir:
The code does detect version conflicts in transitive dependencies.

PR #131, opened 2015-02-17
Add fake classpath show option

jgdavey:
This is useful for IDE authors that may not have an active REPL
connection, but still need static support for navigation, etc.


alandipert:
Cool, thanks!

PR #132, opened 2015-02-18
Atomically update boot-env

dubiousdavid:
Didn't see any tests, so I didn't have any way to formally verify the code.  The current code is not thread safe and is subject to a race condition.  Not sure how much thread safety matters for boot, but I figure better safe than sorry.

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -586,7 +586,7 @@
           v'' (if-let [b (get @cli-base k)] (merge-if-coll b v') v')]
       (assert (printable-readable? v'')
               (format "value not readable by Clojure reader\n%s => %s" (pr-str k) (pr-str v'')))
-      (swap! boot-env update-in [k] (partial pre-env! k) v'' @boot-env))))
+      (swap! boot-env #(update-in % [k] (partial pre-env! k) v'' %)))))
 
 (defn merge-env!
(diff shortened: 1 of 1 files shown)


alandipert:
Thanks!  This is a good catch.  However, I'm inclined to defer addressing thread things until it's clearly a problem in some way.  This is a step in that direction, but I think there's enough going on in core that we don't want to piece-meal it.

Part of the larger issue is that some `pre-env!` methods do side effects, but `pre-env!` might be called multiple times because `(partial pre-env! ...)` happens in the `swap!`.  Really making core thread safe would involve a plan for doing the pre-env! business once and only once before mutating env.

Does this make sense?

alandipert:
I had a little discussion with @micha about this and arrived at the conclusion it's either impossible or extremely hard to coordinate our mutation of the env atom and the environment in the way that we need.  Sometimes the environment mutation (adding deps, for example) changes what we store in the atom, and vice versa.  The best we could do is probably locking.

Still open to cleanup, but there are many subtleties.  As we don't currently face a concrete problem or bug, it's best to defer.

Thanks anyway for the contribution!

dubiousdavid:
That sounds reasonable to me.  Would it make sense to indicate in the documentation that set-env!, merge-env!, etc. are not thread safe?

PR #135, opened 2015-02-20
Suppress updates when running show --fake-classpath

jgdavey:
I missed this before, but the `--fake-classpath` flag was implicitly also printing updates.


micha:
Thanks! :metal:

Issue #189, opened 2015-04-17
Syncing file permissions 


danielsz:
Changing file permissions on files in the fileset don't carry through and "get lost" when those files are written to the file system.

Thoughts for a solution:

When doing a `commit!`, files in the fileset are copied with their attributes.
From an initial source reading, this should happen around [here](https://github.com/boot-clj/boot/blob/11c40541ca51eac7e474f5e03651e0a8f5324555/boot/pod/src/boot/file.clj#L113-L121).

The JVM API to look for could be `Files.copy(Path, Path, CopyOptions)` where `CopyOptions` has `COPY_ATTRIBUTES`.
> The exact file attributes supported are file system and platform dependent, but last-modified-time is supported across platforms and is copied to the target file.

Source: https://docs.oracle.com/javase/tutorial/essential/io/copy.html

micha:
This is actually kind of complicated to implement and has other implications because of the way the fileset works internally. The main problem is that the files underneath the immutable fileset object are themselves immutable and interned, in a way.

When you add files to the fileset the file is copied into a special temp directory, named for the md5 hash of its contents, and write permission removed. This makes the canonical representation of this file immutable. Then, when `commit!` is called on the fileset object, hard links are created from other directories in the working tree to these immutable files.

The problem is that links cannot have different permissions from the things they point to, which means that the canonical representation for a file must now include the permissions as well as the contents. This adds a good deal of complexity in a place where things are already kind of weird.

If anyone has ideas for a clean way to implement this please feel free to reopen this ticket or make a pull request etc.

danielsz:
Thank you for the thorough explanation. This is very helpful.

PR #196, opened 2015-04-24
Copy with attributes

danielsz:
This commit uses `copy` with the `StandardCopyOption` enums.
Related to  #189.

By using the specified options, we can get rid of the following in `copy-with-lastmod`:

* checking if destination exists, and deleting it if it does
* setting last modified attribute.

Additionally, we copy files with other attributes that were previously ignored.


danielsz:
This commit misses the point that the fileset is immutable.

Issue #205, opened 2015-05-26
uber task exclusion regex is too strict


tobias:
The `uber` task currently uses a hard-coded exclusion regex of
`#"^META-INF/"`, which excludes files which may be needed at runtime
from the uberjar (ServiceLoader descriptors, etc). I discussed this
with @alandipert and @micha on IRC, and we came up with a tentative
plan of:

1. Make the default regex more relaxed, similar to [lein's default](https://github.com/technomancy/leiningen/blob/master/leiningen-core/src/leiningen/core/project.clj#L215),
   but also exclude manifests, indexes, and poms (the `jar` task requires
   that only the pom info for the current project be in the
   fileset). Something like:

   ```clojure
   [#"(?i)^META-INF/[^/]*\.(MF|SF|RSA|DSA)$"
    #"(?i)^META-INF/.*pom\.(properties|xml)$"
    #"(?i)^META-INF/INDEX.LIST$"]
   ```

2. Allow the user to override the default with a `--exclude` option,
   and possibly via a `--include` option (though is that more confusing
   than useful?)

3. Allow specifying merging functions via `--merge`. This would allow
merging property files, etc that would collide. This probably needs to
be a map of regex -> fn. Lein does this, but with a [default set of matchers](https://github.com/technomancy/leiningen/tree/master/leiningen-core/src/leiningen/core/project.clj#L215).

tobias:
This issue is just for `--exclude`/`--include`. I've moved `--merge` to #208.

PR #207, opened 2015-05-27
Relax default uber excludes and add options to override exclude and include [#205]

tobias:


micha:
:metal: Looks great, thanks!

PR #213, opened 2015-06-03
Add concat-merger for service providers and add it to standard-jar-mergers

kul:


micha:
@kul Thanks for the PR! Can you please explain the motivation? Fat JARs are tricky; I want to make sure I fully understand what's involved.

micha:
@kul Ah, I see now. Thanks @tobias for the pointer! See [what Leiningen does](https://github.com/technomancy/leiningen/blob/master/leiningen-core/src/leiningen/core/project.clj#L220) and the [ServiceLoader documentation](https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html).

PR #226, opened 2015-06-13
129 add notify task

### LINKED ISSUE #129: Consider adding visual notification task
Visual notifications are a very common feature request for Leiningen plugins (judging from my observation of IRC and Twitter conversations). I haven't seen any requests for audio notifications.

Please consider merging the built in `speak` task with @jeluard's [boot-notify](https://github.com/jeluard/boot-notify) to create a single built in `notify` task with `audible` and `visible` options. Thanks!

RadicalZephyr:
Will fix #129.

I'm submitting this PR early to solicit feedback as I work.  Right now I just have an initial attempt at defining the task options for `notify`.

Currently, I'm assuming that it makes sense to request both audio and visual notifications.


micha:
Hi! Thanks for the PR! Sorry for the delay responding.

Looks good to me. The only reservation I have is adding the `notify` name which could possibly break someone's build. The problem is that built-in tasks are referred into the `boot.user` namespace (i.e. the namespace generated from the `build.boot` file), so if the `build.boot` has code like this in it the build will fail:

```
(require '[foo.bar.baz :refer [notify]])
```

So our options are

- Replace the `speak` task with this implementation, thereby not adding any new names (we need to ensure backwards compatibility with the existing `speak` task of course, but it looks like that's okay).
- We don't care about this situation because the only place where such breakage would cause problems would be in an automated build, and it's unlikely that someone would refer in a name like `notify` in this case.

RadicalZephyr:
I chose `notify` because it seemed like the discussion in issue 129 settled on that as the most sensible name.  And I think that as you said, the potential breakage isn't really a big deal.  Especially since this `notify` task is actually fully backwards compatible with the `notify` task it is based on.

The other consideration is that if we replaced the built-in `speak` task with this, there are some breaking changes that I made to the options that would need to be changed.

So my vote is keeping it named `notify`.

RadicalZephyr:
Okay, this set of changes actually works.  I had to take out the ability for a user to specify their own notifier, so this task is no longer fully compatible with jeluard's version.

If that's a desirable/necessary feature, then I guess we need a way to represent notifiers so they can be passed across pod boundaries.  I'm thinking of a configurable notifier that would take a program name and some maybe a format string for creating the arguments list to be passed to `sh`.

RadicalZephyr:
Hey @micha or @alandipert any chance of further feedback on this?  I'd love to see it get into boot in whatever form :)

alandipert:
Hey thanks for this!  I took another look and it occurs to me that what we really might want is a multimethod dispatching on strings like what `(System/getProperty "os.name")` returns.

Dispatch is currently sealed by the cond and we'd need to separately implement a way for users to specify their own notifier.  With a multimethod, users would have a straightforward way to override or add their own notifiers.

So, what do you think about ditching the deftypes, converting to using a multimethod, and opening up dispatch on the `"os.name"` string?

RadicalZephyr:
Sounds good to me.  I'll take a crack at it, and push the new changes shortly.

RadicalZephyr:
@alandipert Does that look like about what you were expecting?  So then the way a user would set their own notifier to be used would simply be to `(defmethod -notify ...)` for whatever value they want to override?

I'm having trouble getting the build to run so I haven't checked that this actually works correctly yet. I also want to add an example of overriding a new function to the docs as well.

alandipert:
Thanks, this is what I had in mind.  Do you think it's an improvement?

My only other suggestions would be:

* In the Linux and Mac implementations, when `program-exists?` returns false, call `(get-method -notify :default)` instead of `notify-default` directly.
* Name it `notify-method` instead of `-notify`

RadicalZephyr:
Nice, I didn't know about `get-method`!  This structure is definitely more open-closed and extensible.  The one thing I'm still not totally clear on though is how to make it easier/nicer for the user to express their new notifier in their `build.boot` configuration.

Currently, since the actual notifiers are executed inside a worker pod, there would need to be a way for users to pass the code defining their new notifier into the worker pod.  This brings up the question, is it really necessary to run the notification code inside a pod?

alandipert:
Re: pods, that's a great point!  I think it makes sense for pod usage to be Notifier-implementation dependent.  Then, the multi-method can be "top level" and extensible, and Notifier implementors are free to use pods in their implementations at their discretion.

alandipert:
I had a discussion earlier with @micha about this and we had the following ideas:

* The multi-method is fine
* The interface to a task isn't a multimethod, it's the task options.  So, we should an an option to specify a "notification entrypoint" - a qualified symbol pointing to Var that we call as the notification function, if it's specified.  Otherwise, we fall back to the implementations in the multimethod.

Example build.boot:
```
(require '[boot.util :refer [info]])

(defn my-notify [{:keys [message title]}]
  (info (.toUpperCase (format "%s: %s!\n" message title))))

(deftask build
  []
  (notify :notify-fn 'boot.user/my-notify))
```

As before, pods are only necessary if a notify function implementor wants to bring in deps.  The `notify` machinery itself should not run in a pod.

How does all of this sound?

flyboarder:
@alandipert that sounds really interesting, does boot support entrypoints the same way as multi-options?  ie. i can define multiple notifications to fire?

RadicalZephyr:
@alandipert that sounds good to me.  I'll take a pass at it this weekend.  Items for changing are:

- [ ] Add a `notify-fn` option to the `notify` task
- [ ] Move the basic notify machinery out of the worker ns so it doesn't have to be run in a pod
- [ ] Rewire everything so that the current multi-method is the fallback if `notify-fn` is not specified

RadicalZephyr:
@micha @alandipert Sorry for the long hiatus.  Here's a first take at the changes to make the notification method user extensible.

There are still a few other things I want to fix up, but I think the basic structure is looking pretty good now.

RadicalZephyr:
Anymore feedback @alandipert @micha @Deraen?  I'm feeling pretty satisfied with the code right now, and I'd love to see it get merged in :)

micha:
@RadicalZephyr sorry for the long delays...merging finally!

RadicalZephyr:
Awesome!

PR #228, opened 2015-06-14
Fix file handle leaks when unpacking JAR files

ragnard:
Currently, unpacking entries from JAR files is done by first enumerating the entries, constructing JAR entry URLs, like `jar:http://www.foo.com/bar/baz.jar!/COM/foo/Quux.class` and subsequently opening `InputStream`s on these URLs.

However, using URLs to open streams for JAR file entries appears to be problematic.

I got a "Too many open files" exception when using `uber` task and tracked the problem down to the following:

For each file being unpacked from the JAR, a file handle is opened for the JAR file, and for the destination file, however, as the following `strace` output shows, only the destination file handle is closed. The JAR file handle remains open.

```
8924  open("/home/ragge/.m2/repository/cheshire/cheshire/5.4.0/cheshire-5.4.0.jar", O_RDONLY) = 147
8924  open("/nix/store/c00cymmlb0p05wrr7wq1h11yqqd8i75r-oraclejdk-8u45/jre/lib/content-types.properties", O_RDONLY) = 148
8924  close(148)                        = 0
8924  open("/home/ragge/.boot/cache/tmp/home/ragge/projects/github/redacted/redacted/6v2/4naatj/project.clj", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 148
8924  close(148)                        = 0
8924  open("/home/ragge/.m2/repository/cheshire/cheshire/5.4.0/cheshire-5.4.0.jar", O_RDONLY) = 148
8924  open("/home/ragge/.boot/cache/tmp/home/ragge/projects/github/redacted/redacted/6v2/4naatj/META-INF/leiningen/cheshire/cheshire/README.md", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 151
8924  close(151)                        = 0
8924  open("/home/ragge/.m2/repository/cheshire/cheshire/5.4.0/cheshire-5.4.0.jar", O_RDONLY) = 151
8924  open("/home/ragge/.boot/cache/tmp/home/ragge/projects/github/redacted/redacted/6v2/4naatj/META-INF/leiningen/cheshire/cheshire/LICENSE", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 152
8924  close(152)                        = 0
8924  open("/home/ragge/.m2/repository/cheshire/cheshire/5.4.0/cheshire-5.4.0.jar", O_RDONLY) = 152
8924  open("/home/ragge/.boot/cache/tmp/home/ragge/projects/github/redacted/redacted/6v2/4naatj/cheshire/core.clj", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 153
8924  close(153)                        = 0
8924  open("/home/ragge/.m2/repository/cheshire/cheshire/5.4.0/cheshire-5.4.0.jar", O_RDONLY) = 153
8924  open("/home/ragge/.boot/cache/tmp/home/ragge/projects/github/redacted/redacted/6v2/4naatj/cheshire/custom.clj", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 154
8924  close(154)                        = 0
```

This PR uses `JarFile` to explicitly open a jar file and process the entries. This fixes the problem I was seeing, and as a bonus, the JAR file is only opened once, all entries processed, and then closed.

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -103,7 +103,6 @@
             (core/empty-dir! tmp)
             (util/info "Checking out %s...\n" path)
             (pod/unpack-jar (.getPath file) tmp
-              :cache false
               :exclude pod/standard-jar-exclusions)))
         (->> tmps vals (reduce core/add-source fileset) core/commit!)))))
 
--- a/boot/pod/src/boot/pod.clj
+++ b/boot/pod/src/boot/pod.clj
@@ -298,11 +298,14 @@
           path (.getPath f)]
       (when (.endsWith path ".jar")
         (when (and (.exists f) (.isFile f))
-          (->> f JarFile. .entries enumeration-seq
-               (keep #(when-not (.isDirectory %)
-                        (let [name (.getName %)]
-                          [name (->> (io/file (io/file (str path "!")) name)
-                                     .toURI .toURL .toString (str "jar:"))])))))))))
+          (with-open [jf (JarFile. f)]
+            (->> (.entries jf)
+                 enumeration-seq
+                 (keep #(when-not (.isDirectory %)
+                          (let [name (.getName %)]
+                            [name (->> (io/file (io/file (str path "!")) name)
+                                       .toURI .toURL .toString (str "jar:"))])))
+                 (into []))))))))
 
 (def jar-entries-memoized* (memoize jar-entries*))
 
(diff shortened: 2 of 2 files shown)


ragnard:
This PR means that the `jar-entries` fn is no longer used by boot, but I wasn't sure if it was considered part of public API or not so left it in there. Same goes for `copy-url`.

boot/core/src/boot/task/built_in.clj
```diff
@@ -103,7 +103,6 @@
             (core/empty-dir! tmp)
             (util/info "Checking out %s...\n" path)
             (pod/unpack-jar (.getPath file) tmp
-              :cache false
```
micha: We want to prevent caching of JAR entries here. If they're cached the checkout task won't see changes to the contents of the JAR, which defeats the purpose of checkouts :)

boot/pod/src/boot/pod.clj
```diff
@@ -339,37 +342,42 @@
    [#".*"                   first-wins-merger]])
 
 (defn merge-duplicate-jar-entry
-  [mergers curr-file new-path new-url & {:keys [merged-url]}]
+  [mergers curr-file new-path new-stream & {:keys [merged-url]}]
```
micha: Does this break [the JAR entry &rarr; fileset uber merging code](https://github.com/boot-clj/boot/blob/09eadb04a55e3602daf553d72626f57860f49cb8/boot/pod/src/boot/tmpdir.clj#L84)?

ragnard: Whoopsie. Almost certainly. Will look into that.

boot/core/src/boot/task/built_in.clj
ragnard: AFAICT, the only places `:cache` is currently used is in:

- `jar-entries`  where it toggled the use of a memoized version of the function or not (https://github.com/boot-clj/boot/blob/master/boot/pod/src/boot/pod.clj#L313)
- `copy-url` where it was used to `setUseCaches` on the URLConnection before creating a stream

Neither of these functions are used by `unpack-jar` anymore.

micha: Ah, I see. So the bizarre caching behavior of the `URLConnection` isn't an issue anymore because we're not using `URLConnection`s anymore, right?

:metal:

PR #234, opened 2015-06-21
Add path exclude option to aot task.

kul:
This pr adds exclude option to `aot` task. Although it seems obvious at first to add exclude for namespaces but adding exclusion for path would provide more flexibility in terms of duplicate file names.

By default exclude default to data_reader exclusion regex.

Also see #231 .


micha:
@kul I had completely forgotten about `clojure.tools.namespace.find`! I think ctnf has some nice advantages over exclusions, namely less config and automatic support for .cljc files. What do you think? Feel free to reopen this PR if you like :)

kul:
@micha nice one! :+1:

PR #236, opened 2015-06-24
Modify javac task to enable passing javac options.

kul:
Enables something like `(task/javac :options ["-source" "1.6" "-target" "1.6"])`

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -548,7 +548,7 @@
 
 (core/deftask javac
   "Compile java sources."
-  []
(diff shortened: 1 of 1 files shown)


alandipert:
Awesome, thanks!

Can we go with an options map instead though?  Like this:
` [o options KEY=VALUE {str str} "Options passed to the Java compiler."]`

kul:
Sure we can do that, but some of the javac options would have no values e.g. -verbose -nowarn etc. WDYT?

alandipert:
Ah I forgot about those - probably better to just stick with a list of strings then.

PR #239, opened 2015-06-30
Add credentials options to `push` task.

kul:
This allows adding credentials on-the-fly from a given file path. The credentials file contains a clojure map and has keys either exact url string or regex that finds itself in repo url. e.g.

```clojure
{#"clojars" {:username "foo" :password "bar"}
 #"aws-private-repo-bucket" {:username "...." :passphrase "..."}
 "http://somerepo/release" {:username "...." :password "..."}}
```

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -688,6 +688,7 @@
 
   [f file PATH            str      "The jar file to deploy."
    F file-regex MATCH     #{regex} "The set of regexes of paths to deploy."
+   c credentials PATH     str      "The path to credentials file for repositores."
    g gpg-sign             bool     "Sign jar using GPG private key."
    k gpg-user-id NAME     str      "The name used to find the GPG key."
    K gpg-keyring PATH     str      "The path to secring.gpg file to use for signing."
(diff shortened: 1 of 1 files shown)


micha:
Hi @kul sorry for the delayed response. This looks great, thanks! There is an `edn` argument type that we could use to parse the new argument, but it's not a big deal. If you want to update to use that it's cool, or we can just merge as is.

kul:
@micha thanks for review. Right now the `cred` option in the task is a path to file which would contain the edn. Do you mean to modify it such that it can accept both path to edn file as well as edn itself?

kul:
Hi @micha can we merge this as making `cred` accept file path as well as edn would be tricky. Moreover i dont see any use of making it accept edn as user would copy it from some file so why not just give the file path directly. WDYT?

alandipert:
@kul thanks for the contribution!  After considering this again for 2.2.0, we decided to pass on this, because it seems a little too special-casey to be in the general-as-possible push task.

For global repo configuration, users could add repo config in a manner similar to Maven's `settings.xml` by putting a `(set-env! :repositories ...)` call in ~/$BOOT_HOME/profile.boot`.

For project-local config, users could do e.g.:
```
(set-env! :repositories (read-string (slurp "repos.edn")))
```
or they could make a short task to do the same.

I propose we leave this ticket open to attract future users.  If it seems to attract more attention, then we'll know it could be general enough to go in `push` or for us to otherwise devise a solution.

Does that sound OK?

kul:
@alandipert `(set-env! :repositories ...)` makes perfect sense. Thanks for going through it.

Deraen:
This is effectively implemented using `:repo-map` option on https://github.com/boot-clj/boot/pull/311

Issue #243, opened 2015-07-08
Can't use BOOT_LOCAL_REPO on windows


schmir:
I have set the BOOT_LOCAL_REPO variable on windows. Running boot repl
now fails with a FileNotFoundException:

    C:\Users\ralf\home>set BOOT_HOME
    BOOT_HOME=C:\home\.boot

    C:\Users\ralf\home>set BOOT_LOCAL_REPO
    BOOT_LOCAL_REPO=C:\home\.m2\repository

    C:\Users\ralf\home>boot repl
    Exception in thread "main" java.io.FileNotFoundException: C:\home\.boot\cache\boot\customC:\home\.m2\repository\1.6.0\2.1.2\deps.cache (The filename, directory name, or volume label syntax is incorrect)
	    at java.io.FileOutputStream.open0(Native Method)
	    at java.io.FileOutputStream.open(FileOutputStream.java:270)
	    at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	    at java.io.FileOutputStream.<init>(FileOutputStream.java:162)
	    at boot.App.writeCache(App.java:123)
	    at boot.App.readCache(App.java:137)
	    at boot.App.main(App.java:316)

The problem is that boot tries to use the local repository path as a
part of another filename [1].

[1] https://github.com/boot-clj/boot/blob/master/boot/base/src/main/java/boot/App.java#L285

PR #244, opened 2015-07-08
make environment variable BOOT_LOCAL_REPO usable on windows

schmir:
replace special characters when building path components when the
environment variable BOOT_LOCAL_REPO is set.
Previously we ended up trying to use illegal filenames like
C:\home\.boot\cache\boot\customC:\home\.m2\... on windows.

Since we now also replace the path separators, we use a single directory
under boot/custom/ instead of a deep directory structure.

This may lead to a single path component being too long. But that is a
future issue.

see https://github.com/boot-clj/boot/issues/243 for more information.


schmir:
superseded by #246

PR #246, opened 2015-07-09
make environment variable BOOT_LOCAL_REPO usable on windows

schmir:
use md5 hash when building path components when the environment variable
BOOT_LOCAL_REPO is set.

Previously we ended up trying to use illegal filenames like
C:\home\.boot\cache\boot\customC:\home\.m2\... on windows.

We now use a single directory under boot/custom/ instead of a deep
directory structure.

The conversion to a hex string is a bit awkward, see
https://stackoverflow.com/questions/9655181/how-to-convert-a-byte-array-to-a-hex-string-in-java
for a discussion.

fixes https://github.com/boot-clj/boot/issues/243


micha:
:+1: Awesome!

PR #261, opened 2015-08-04
Add CHANGES.md with change log since 2.1.2

ragnard:
Draft of a change log since 2.1.2 derived from git log. 

Would be nice to make issues/commits referenced proper links, can look into that if the general content and idea of a change log file is deemed good.


micha:
Thanks @ragnard this is great!

Issue #238, opened 2015-06-27
boot.sh for newer versions?


donbonifacio:
The install instructions and travis config I've seen use `boot.sh`, but that isn't present, for example, on release `2.1.2`.

micha:
This is because there have been no changes to `boot.sh` since version `2.0.0`

donbonifacio:
How can I install  a more recent boot? I'm under the impression that `boot.sh` installs `2.0.0`, and that I need a newer version to use reader conditionals. Don't know if I did something wrong, I ended up falling back to lein for multi-platform builds.

micha:
If you get boot.sh from the 2.0.0 github release download you will have the latest version of the boot jar as of now. However Clojure 1.7.0 was just released last week so it's not yet the default for boot. To make 1.7.0 the default in boot you can do:

```
$ BOOT_CLOJURE_VERSION=1.7.0 boot -u
```

Once you're using Clojure 1.7.0 you will be able to use .cljc files and reader conditionals.

heypiotr:
Is this still true? (I know there hasn't been any new releases since the 2.1.2 mentioned in the original post, so I assume it is.)

I'm not entirely sure I understand why boot.sh hasn't changed since 2.0.0, even when there have been changes to the Boot's source in the meantime—do you mind explaining this for newbies like me?

martinklepsch:
@heypiotr the `boot.sh` script is just a very thin shim on top of Maven to load "the real Boot". This part really just bootstraps the environment and most of the functionality Boot provides comes from the jars that you can easily update with `boot -u`. Because it doesn't do a whole lot it's also changed much less frequently. 

Does that make sense?

heypiotr:
Totally, thank you! :rocket:

(May it'd be a good idea to add this to the installations section of the README?)

martinklepsch:
@heypiotr Yes, adding two lines below the instructions might be good to clear things up. Wanna do it? :smiley:

heypiotr:
Here you go. Shamelessly ripped off some of your explanation, hope that's okay! :blush:

PR #262, opened 2015-08-04
Update README.md

heypiotr:
The goal is to explain the difference between the version of the wrapper
and the "real Boot," i.e., answer the "why is the latest version of Boot
available for download 2.0, while I can see the actual latest is 2.1.2"
questions.

Should fix https://github.com/boot-clj/boot/issues/238 and the likes of
it.


alandipert:
Thanks!

PR #274, opened 2015-08-18
New implementation for signing releases

danielsz:
* Client libraries like bootlaces should not have to configure anything.
* Signing and encrypting in Boot-clj works out of the box, with no configuration. This can be made to work because we are going to make assumptions about the end user's environment. In turn, those defaults will be capable of being overridden with environment variables.
* Secret keyring location is the default one ($HOME/.gnupg/secring.pgp). Can be overridden with BOOT_GPG_SECRING.
* Name/Email is the one we find. Provided the secret keyring has only one user, we can be sure that we have the correct one. If more than one are found, we throw an error. Name/Email can be set with BOOT_GPG_USER.
* Default signing key is the first one returned by (keyring/list-public-keys keyring). This will work on a straightforward GPG setup. If the user follows best practices and has derived subkeys from his master keypair, he can set those with BOOT_GPG_SIGNING_KEY_ID and BOOT_GPG_ENCRYPTING_KEY_ID.

### DIFF
--- a/boot/core/src/boot/task_helpers.clj
+++ b/boot/core/src/boot/task_helpers.clj
@@ -35,21 +35,11 @@
         (->> sym ns-aliases (into base) (mapcat pubs)))
       (filter task?) (sort-by :name) (group-by :ns*) (into (sorted-map)))))
 
-(defn read-pass
-  [prompt]
-  (String/valueOf (.readPassword (System/console) prompt nil)))
-
-(defn sign-jar [out jar pass keyring user-id]
-  (let [prompt (pod/with-call-worker
-                 (boot.pgp/prompt-for ~keyring ~user-id))
-        pass   (or pass (read-pass prompt))]
-    (pod/with-call-worker
-      (boot.pgp/sign-jar
-        ~(.getPath out)
-        ~(.getPath jar)
-        ~pass
-        :keyring ~keyring
-        :user-id ~user-id))))
+(defn sign-jar [out jar]
+  (pod/with-call-worker
+    (boot.pgp/sign-jar
+     ~(.getPath out)
+     ~(.getPath jar))))
 
 (defn print-fileset
   [fileset]
--- a/boot/worker/src/boot/pgp.clj
+++ b/boot/worker/src/boot/pgp.clj
@@ -2,56 +2,74 @@
   (:require
    [clojure.java.io   :as io]
    [boot.pod          :as pod]
-   [mvxcvi.crypto.pgp :as pgp])
+   (clj-pgp
+     [core :as pgp]
+     [keyring :as keyring]
+     [signature :as pgp-sig]))
   (:import
    [java.util.regex Pattern]
    [java.io StringBufferInputStream]))
 
-(def default-secring
-  (-> (System/getProperty "user.home")
-    (io/file ".gnupg" "secring.gpg")))
+(defn find-secring []
+  (let [home (System/getenv "HOME")
+        locations [(System/getenv "GNUPGHOME") (System/getenv "BOOT_GPG_SECRING") (str home "/.gnupg/secring.gpg") (str home "/.gpg/secring.gpg")]
+        exists-fn #(when (and % (.exists (io/as-file %))) %)
+        secring (some exists-fn locations)]
+    (or secring (throw (Exception. "Can't find secring file. Please set 'BOOT_GPG_SECRING'")))))
 
-(defn secring
-  [& [file]]
-  (pgp/load-secret-keyring (io/file (or file default-secring))))
+(defn get-keyring []
+  (keyring/load-secret-keyring (io/file (find-secring))))
 
-(defn get-pubkey
-  [secring & [user]]
-  (let [ks      (pgp/list-public-keys secring)
-        re-user #(re-find (re-pattern (Pattern/quote (str user))) %)
-        match?  #(some re-user (-> % pgp/key-info :user-ids))]
-    (let [ks' (if-not user ks (filter match? ks))]
-      (if (or (not user) (<= (count ks') 1))
-        (first ks')
-        (throw (ex-info "multiple keys match"
-                 {:keys (mapv pgp/key-info ks')}))))))
+(defn find-master-key-by-user [user]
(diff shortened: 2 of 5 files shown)


boot/core/src/boot/task_helpers.clj
```diff
@@ -35,21 +35,11 @@
         (->> sym ns-aliases (into base) (mapcat pubs)))
       (filter task?) (sort-by :name) (group-by :ns*) (into (sorted-map)))))
 
-(defn read-pass
-  [prompt]
-  (String/valueOf (.readPassword (System/console) prompt nil)))
-
```
micha: We may want to either keep this or move it to `boot.util`? It seems like a handy function to have hanging around.

martinklepsch: @micha it's still in `boot.pgp` but I agree that a fn reading user input could be in `boot.util` as well.

version.properties
```diff
@@ -1 +1 @@
-version=2.2.0
+version=2.2.1
```
micha: This is a breaking change to the API, no? We've removed options from tasks, etc. Perhaps we bump to 3.0.0?

boot/worker/src/boot/pgp.clj
```diff
@@ -2,56 +2,74 @@
...
+  (let [keyring (get-keyring)]
+    (if-let [hex-id (System/getenv "BOOT_GPG_SIGNING_KEY_ID")]
+      (get-private-key-by-id hex-id passphrase)
+      (-> (find-master-key)
+          pgp/key-info
+          :key-id
+          (#(keyring/get-secret-key keyring %))
+          (pgp/unlock-key passphrase)))))
+
+(defn read-pass
```
frankiesardo: I'm concerned about having a prompt as the only way to pass a passphrase around. Unattended builds (CIs) need another option to be able to run. Maybe it's possible to use the .gpgconf? I know you can specify `passphrase=..` in it.

danielsz: The current approach was to use a Java API as opposed to shelling out to the `gpg` binary, so `.gpgconf` does not apply. 
Generally speaking, using `passphrase` in `.gpgconf` should be avoided. Passing the passphrase automatically can be achieved with `gpg-agent`. But this also relies on    the `gpg` binary.

danielsz: A straightforward solution (but not safe) solution to your concern would be to add a BOOT_GPG_PASSPHRASE environment variable that, if set, will be used instead of the prompt.

frankiesardo: I know, lein uses `gpg-agent` which I quite like as an approach because it shifts complexity on the tool itself that you can configure using `.gpgconf`. I'd rather have `push` accept optionally the passphrase so in my CI machines I can have a `profile.boot` that does `(task-options! push {:passphrase "My pass"})`. Environmental variable might be overkill in this case.

frankiesardo: Re: environmental variables adding another one might be more in line with the implementation of this pull request, but I'm actually arguing it would be better to remove all of them and pass them explicitly as optional arguments (making the configuration explicit when you call the function rather than relying on the global environment state). Just my 2 cents.

danielsz: In the current PR, GPG configuration is driven by environment variables. They are documented here: https://github.com/boot-clj/boot/wiki/Deploying-with-Boot
For consistency's sake, the passphrase feature would be just an additional environment variable. Furthermore, the advantage of an environment variable is that the user can use a third-party library like `bootlaces` to deploy artefacts. He doesn't need to deal directly with the `push` task, and yet he still retains control over configuration options.

frankiesardo: I think `He doesn't need to deal directly with the push task, and yet he still retains control over configuration options.` is a non-argument. Rather than having my configuration nicely in a profile.boot now I need to export an environment variable.

Each environment variable you add to the project you add complexity and clutter. If it's only used by one task, then why not it be a parameter to the task itself? You get automatic documentation for it and you can set it programmatically rather than shelling out.

I mean, fine by all means if you want to give the option to drive the implementation via env vars, but don't take away the task options: they are much more easier to use and repl-friendly.

danielsz: I have no problem with whatever configuration mechanism we are going to adopt. This PR is a proof of concept that can be refined and finalized by others. @micha has started this   https://github.com/adzerk-oss/env which will likely address this problem. Please feel free to contribute to this PR.

danielsz:
@micha @alandipert Bumping this PR. Is there something I can do to help get this accepted. I would love to hear your comments or reservations. If I remember correctly, this PR doesn't break anything.

Deraen:
The implementation looks good but I feel strongly that `gpg-agent` should be used instead of password prompt.

Deraen:
Is there a reason for not using `gpg` binary for everything? Leiningen is doing it and I don't see any problems with it. In this case I think it would be the best to do what everyone else is doing.

micha:
I didn't use the `gpg` binary mainly for windows support. Also I always have trouble with configuring Leiningen with GPG (notice boot itself isn't signed because I can't get it to work reliably).

Deraen:
Leiningen docs mention GPG4Win: https://github.com/technomancy/leiningen/blob/master/doc/GPG.md#windows

micha:
I'm on board with using gpg, especially gpg-agent. Since it uses unix domain sockets we can't do it from java, unfortunately, and will have to shell out to gpg.

Deraen:
I did quick test with `gpg` binary using code from Leiningen: https://github.com/Deraen/boot-contrib-gpg/blob/master/src/deraen/boot_contrib_gpg.clj

In addition to signing, it supports reading repository credentials from `~/.boot/credentials.edn.gpg`.

danielsz:
Awesome. Any implementation solving the limitations of the current one is a big win. Leveraging `gpg-agent` for configuration-less signing and reading credentials is also a big win.

Deraen:
I merged the alternative implementation using GPG binary.

PR #309, opened 2015-10-10
no javac on empty fileset

instilled:
Avoid javac invocation on empty fileset and clean up some whitespaces checkout task.


micha:
Looks great, thanks! :+1:

PR #311, opened 2015-10-18
Use gpg binary for signing and reading credentials from encrypted file

Deraen:
Copied the implementation from Leiningen.

- Adds new `:deploy-repositories` env property. This is because it's
probable that credentials between repositories differ for read and
deploy.

Needs still some polishing.


boot/aether/src/boot/aether.clj
```diff
@@ -59,13 +64,82 @@
           :password        password
           :non-proxy-hosts (get-non-proxy-hosts)}))))
 
+(defn ^{:boot/from :technomancy/leiningen} boot-home
```
Deraen: Is there already somewhere where this is available? `App` is at least reading the env variable.

boot/aether/src/boot/aether.clj
```diff
@@ -59,13 +64,82 @@
...
+                  :else v))]
+    (if (#{:username :password :passphrase :private-key-file} k)
+      (assoc result k (resolve v))
+      (assoc result k v))))
+
+(defn ^{:boot/from :technomancy/leiningen} resolve-credentials
+  "Applies credentials from the environment or ~/.boot/credentials.edn.gpg
+  as they are specified and available."
+  [settings]
+  (let [gpg-creds (if (= :gpg (:creds settings))
```
Deraen: Not sure if `:creds` is useful. AFAIK `{:url "..." :creds :gpg}` is identical to `{:url "..." :username :gpg :password :gpg}`.

Perhaps only useful for Leiningen compatibility.

boot/aether/src/boot/aether.clj
```diff
@@ -59,13 +64,82 @@
...
+      (dissoc (merge gpg-creds resolved) :creds)
+      resolved)))
+
 (defn resolve-dependencies*
   [env]
   (try
     (aether/resolve-dependencies
       :coordinates       (:dependencies env)
       :repositories      (->> (or (:repositories env) @default-repositories)
-                           (map (juxt first (fn [[x y]] (if (map? y) y {:url y}))))
```
Deraen: I dropped suport for defining repositories as string instead of map with `:url` property. Could be still added back.

Deraen:
This is alternative to #274 

Btw. reading credentials a file or other sources could be provided as a separate library, but I think deploy repositories should be separated in any case.

Deraen:
Ah, I almost forgot. There's a reason why it would be preferred to have support for reading credentials from file in Boot itself. With a plugin the use would look something like:

```clj
(set-env! :dependencies '[[deraen/boot-contrib-gpg "0.1.0-SNAPSHOT"]])
(require '[deraen.boot-contrib-gpg :refer [set-repositories! push-gpg]])

(set-repositories! [["my.datomic.com" {:url "https://my.datomic.com/repo"
                                       :creds :gpg}]])

(set-env! :dependencies '[rest of dependencies, including stuff from private repos])
```

It's quite verbose because the plugin has to be added before adding other dependencies.

Deraen:
- All existing options should be kept for compatibility
- If password is provided for deploy task, the password should be used at least for signing
    - Could either pass the password to GPG binary or use current implementation?
    - Keeping current implementation could be useful for cases where GPG isn't available?
- Reading repository credentials is a new feature but it should probably also work with existing GPG options?
- Deploy repositories should be moved to deploy task options

Deraen:
Looks like it would be possible to extend Bouncy Castles PGP library with PassphraseLoader which would retrieve key from gpg-agent: https://github.com/kohsuke/pgp-maven-plugin/blob/master/src/main/java/org/kohsuke/maven/pgp/loaders/GpgAgentPassPhraseLoader.java

Would only work for Unix systems but Windows users could anyway use the task options.

Deraen:
This should now support all existing options, though `gpg-keyring` is deprecated.

@danielsz What was the use case for providing GPG options through environment variables in another PR? Should I implement that also, or is using gpg binary enough?

danielsz:
@Deraen  Off the top of my head, the environment variables are redundant if gpg-agent is used, because it knows who the user is, what the signing key is, etc...
In the other PR, all this was "guessed", and the environment variables were meant to override the "guessing".

boot/aether/src/boot/aether.clj
Deraen: Once #317 is merged, `boot.App/bootdir` should probably be used.

Deraen:
Rebased and updated to use `App/bootdir`.

Deraen:
And squashed.

Issue #248, opened 2015-07-10
Spurious FSEventStreamStart log messages on OS X


aaronc:
See this thread for details: https://groups.google.com/forum/#!topic/clojurescript/KUcELEvX18A

alandipert:
Can you please provide the OS X point version and output of `boot -V`?

aaronc:
OS X 10.10.3 (14D136)

```
#https://github.com/boot-clj/boot
#Fri Jul 10 21:42:45 EDT 2015
BOOT_CLOJURE_VERSION=1.7.0
BOOT_VERSION=2.1.2
#App version: 2.0.0
```

There is a pending update to 10.10.4 from the App Store - maybe I should install that.

alandipert:
@aaronc Thanks, and yes if you can test on 10.10.4 that would be helpful as well.

drakezhard:
I have the same problem with the same boot version and OS X version. The update to 10.10.4 didn't fix it.

kirked:
Replied in the group thread, but I found [this PR](https://github.com/joyent/libuv/pull/894) for libuv from 2013 when they had the issue, and an [older post](https://lists.apple.com/archives/darwin-dev/2008/Feb/msg00043.html) that describe the problem.

Apparently there are too many filesystem-change watch processes with large file sets on the same device. The fix implemented in libuv appears to be to use a single watcher for a device and filter the changes on notification to the ones you're really interested in.

drakezhard:
Yeah I figured it out myself just now, I added "target" to resources to serve files properly with pedestal, and triggered an infinite watch loop. (set-env! :resource-paths #(disj % "target")) before the development task fixed it. Thanks anyway kirked.

piranha:
Same thing for me when I'm building uberjar! Though removing "target" from `:resource-paths` doesn't really help, I'm pointing to different `:target-path` when building uberjar now.

cigitia:
I’m encountering the same problem for all boot tasks, including `boot` only and `boot repl`. No tasks are able to run at all. The error’s output is identical with that in the Google Groups thread.

Restarting my computer does not resolve the problem; clearing the `build.boot` file does not resolve the problem. I was able temporarily to resolve the problem by clearing all files from `~/.boot/tmp` and clearing my `build.boot` file to have blank content, but restoring my old `build.boot` file made the problem reoccur. 

I'm using OS X 10.11 (15A284). `boot -v` outputs:

    #https://github.com/boot-clj/boot
[...]

PR #312, opened 2015-10-22
allow exclusion of specific directories from being watched

martinklepsch:
useful for node_modules and similar stuff

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -239,11 +239,21 @@
    new))
 
 (defn- order-set-env-keys
-  "Ensures that :dependencies are processed last, because changes to other
-  keys affect dependency resolution."
+  "Ensures that
+    - :dependencies are processed last, because changes to other
+      keys affect dependency resolution.
+    - :watcher-debounce and :watcher-exclude settings are processed
+      first because otherwise they're not used when loading input dirs."
   [kvs]
-  (let [dk :dependencies]
-    (->> kvs (sort-by first #(cond (= %1 dk) 1 (= %2 dk) -1 :else 0)))))
+  (let [dk       :dependencies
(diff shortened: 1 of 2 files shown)


martinklepsch:
@micha re your comment on #248

That would be less predictable and I assume the implementation would be more complex.
- We would need to traverse the tree and count files
- Then decide where we cut it off. Don't see a good decision criteria for this yet.

micha:
We already are traversing the tree when we add the watchers. The way the fsevent stuff works you need to add a watcher to each directory---it needs a separate one for each directory. Like if you have a directory with subdirectories you need to add a watcher to each of the subdirectories and to each of their subdirectories etc. This is already in place, so we could do it there, couldn't we?

micha:
Also boot relies on the file watchers to function correctly if you want those files to be in the fileset. So for something like node_modules i think you'd want a task that can just sync them to the target without ingesting them into the fileset. This will be much easier if we eliminated automatic target syncing and made it a separate task, as we discussed on slack.

martinklepsch:
> Also boot relies on the file watchers to function correctly if you want those files to be in the fileset. 
The files in excluded dirs were still present in the fileset. I was also somewhat surprised tbh.

> So for something like node_modules i think you'd want a task that can just sync them to the target without ingesting them into the fileset.
Could I also add them to the classpath in a "static" way?

---
The traversing is there but essentially we would need another full traversal to create a tree of directories and filecounts right? Then we would still need to come up with some potentially arbitrary measures to decide whether to cut of a subtree from watching.

martinklepsch:
> So for something like node_modules i think you'd want a task that can just sync them to the target without ingesting them into the fileset.

You might also want them on the classpath — are there means to put stuff on the classpath without it going through the fileset? Maybe the solution is a task that packages `node_modules` as a jar? 

Then we can also just forget about all of this stuff.

Issue #268, opened 2015-08-07
Classes loaded by Pods are leaked


stephenbrady:
This repo produces a class leak with each invocation of `test` by `boot watch`.

https://github.com/stephenbrady/boot-pod-leak

In writing custom boot tasks, I see the same kind of class leak when using either `boot.pod/make-pod` with `boot.pod/destroy-pod` or `boot.pod/pod-pool` with `:refresh`.

martinklepsch:
Hey @stephenbrady thanks for the report!

How exactly did you detect this memory leak? Did you get `java.lang.OutOfMemoryError: Java heap space` when your heap filled up or did you get `java.lang.OutOfMemoryError: PermGen space` errors?

stephenbrady:
Sure. By profiling in jvisualvm. I can see with each invocation of `test` that the number of classes in memory grows by roughly the number of new classes you would expect to see with each new loading of the clojure library. The biggest symptom is how long it takes to run each marginal invocation of `test` as the number of classes just grows and slows things down in the jvm.

This is with jdk8, so no perm gem memory issues.

martinklepsch:
@stephenbrady Thanks, that confirms my understanding.

martinklepsch:
@alandipert:
> @martinklepsch, @micha re: mem core.async might be a culprit, https://github.com/projectodd/shimdandy#preventing-memory-leaks

I read that page but assumed that Boot does not use/depend on core.async and searching for classes in the YourKit classbrowser doesn't list any classes containing "async".

The program I use for testing also is very minimal and does not load any additional deps:
```clj
#!/usr/bin/env boot

(require '[boot.pod :as pod])

(defn -main [& args]
  (loop [i 0]
    (let [c-pod (pod/make-pod (get-env))]
      (boot.util/info "New pod taken: %s\n"  c-pod)
      (pod/with-eval-in c-pod
        (boot.util/info "Hi from pod no. %s.\n" ~i))
      (pod/destroy-pod c-pod)
      (recur (inc i)))))
```

tobias:
I believe I have figured out what is causing this - I believe boot is
triggering a call to either `clojure.lang.Var.cloneThreadBindingFrame`
(which will happen if `clojure.core/binding-conveyor-fn` is called,
either via a call to `clojure.core/future` or one of the
`clojure.core/send*` fns), or `clojure.lang.Var.getThreadBindings`
(which will happen if `clojure.core/get-thread-bindings` is called,
[...]

PR #314, opened 2015-10-23
Update to ShimDandy 1.2.0

tobias:
1.2.0 removes an avenue for leaking memory, and provides a close method
to call `shutdown-agents` on your behalf.

This should address #268.

### DIFF
--- a/boot/base/pom.in.xml
+++ b/boot/base/pom.in.xml
@@ -86,7 +86,7 @@
     <dependency>
       <groupId>org.projectodd.shimdandy</groupId>
       <artifactId>shimdandy-api</artifactId>
-      <version>1.1.0</version>
+      <version>1.2.0</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
--- a/boot/pod/project.clj
+++ b/boot/pod/project.clj
@@ -18,4 +18,4 @@
   :dependencies [[boot/base                               ~version :scope "provided"]
                  [org.clojure/clojure                     "1.6.0"  :scope "provided"]
                  [org.tcrawley/dynapath                   "0.2.3"  :scope "compile"]
-                 [org.projectodd.shimdandy/shimdandy-impl "1.1.0"  :scope "compile"]])
+                 [org.projectodd.shimdandy/shimdandy-impl "1.2.0"  :scope "compile"]])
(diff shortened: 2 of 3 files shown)


martinklepsch:
:+1: I'll give this a test drive later today and report back/merge. Thanks so much for your help on this @tobias!

martinklepsch:
@micha @alandipert @Deraen this requires a new binary release, did we come up with a consistent scheme for versioning those?

Deraen:
@martinklepsch In this case binary change is not breaking but will just fix problems? So I think version 2.4.0 would be good enough.

Deraen:
I could be wrong about breakage, the `close` method is new so calling it will break with old shimdandy.

martinklepsch:
@Deraen it's breaking as in old binary causes failure when ran with new libs:
```
        clojure.lang.ExceptionInfo: No matching field found: close for class org.projectodd.shimdandy.impl.ClojureRuntimeShimImpl
    data: {:file
           "/var/folders/ss/4qg3hk1d4nv40phg1360ng5w0000gn/T/boot.user4226953933120525930.clj",
           :line 11}
java.lang.IllegalArgumentException: No matching field found: close for class org.projectodd.shimdandy.impl.ClojureRuntimeShimImpl
                 ...
boot.pod/destroy-pod                           pod.clj: 470
     boot.user/-main  boot.user4226953933120525930.clj:   7
                 ...
   boot.user/eval243  boot.user4226953933120525930.clj:  11
                 ...
  boot.main/-main/fn                          main.clj: 161
     boot.main/-main                          main.clj: 161
                 ...
    boot.App.runBoot                          App.java: 242
       boot.App.main                          App.java: 356
```

micha:
I think we should just wrap with try/catch:

```clojure
(when pod
  (try (.close pod)
    (catch java.lang.IllegalArgumentException _
      (.invoke pod "clojure.core/shutdown-agents")
      (.. pod getClassLoader close))))
```

This will fall back to the previous behavior if an older version if shimdandy is present.

tobias:
Ah, I neglected to consider how this would work with the app/lib split. It would also be safe to not call `.close` at all, and just continue to call `shutdown-agents` yourself, as that is all `.close` does currently.

I should also clarify that since the changes are all in shimdandy-impl, and the api changes were only additive (just `.close`), then a lib release alone will fix the memory leak, no need to rev the app.

micha:
Awesome! Thanks @tobias @martinklepsch @Deraen! :beers:

PR #318, opened 2015-10-26
Fileset commit! performance improvements via patch operations

martinklepsch:
This PR improves the performance of fileset `commit!` operations by storing previously committed state and generating patch operations by diffing previous states and the to-be-committed fileset.


martinklepsch:
@micha should we close this? Another PR with tests only?

arichiardi:
@martinklepsch my gut feeling is that you can have a look at #401 and yes add a PR with the tests whenever that's in.

PR #327, opened 2015-10-28
throw exception when mv source does not exist; fixes #325

### LINKED ISSUE #325: Exeption not thrown when from-path does not exist in fileset
`if-let` below will always be true.
```clojure
  (mv [this from-path to-path]
    (if (= from-path to-path)
      this
      (if-let [f (assoc (get-in this [:tree from-path]) :path to-path)]
        (update-in this [:tree] #(-> % (assoc to-path f) (dissoc from-path)))
        (throw (Exception. (format "not in fileset (%s)" from-path))))))
```

martinklepsch:
fixes #325

### DIFF
--- a/boot/pod/src/boot/tmpdir.clj
+++ b/boot/pod/src/boot/tmpdir.clj
@@ -95,8 +95,9 @@
   (mv [this from-path to-path]
     (if (= from-path to-path)
       this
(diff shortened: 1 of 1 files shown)


micha:
:+1: looks good to me!

PR #329, opened 2015-10-28
Build Boot with Boot

martinklepsch:
It's real! :tada: :confetti_ball: 

- `make install` can still be used to install everything locally.

- `boot build (--watch) (--install)` can be used during development. It will start multiple boot pipelines to build boot libraries when relevant files change. It **is required** to run `make install` before doing this*.  
  >  **Reason for needing to run `make install` once:** it copies the aether uberjar into `boot/base/resources` where it will be picked up when building the base uberjar (`boot.jar`). If the aether uberjar is not present building the base uberjar will fail.
- `make deploy` uses the regular `push-release` provided by bootlaces. I'm not sure how to test this without making a release or creating garbage on clojars, any ideas? Also #328 would be needed.
- there was a `:jar-exclusions [#"^clojure/core/"]` in all `project.clj`s but I don't think that had any effect. Leiningen provides [`:uberjar-exclusions`](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L388) for uberjars where it would make sense but we don't use that.

**Limitation**: We cannot install the version of boot we're currently running. This "will pull the carpet under the feet" of the running process. In practice this isn't a big problem I think but worth mentioning.

I can squash and delete `project.clj` and some other cruft when everything seems alright. :beers:


arichiardi:
Is this patch working still valid? It is really handy when you work on boot stuff...

martinklepsch:
@arichiardi #340 has all the changes in here and is better (don't exactly remember diff anymore). 

Both PRs however have a the problem of not allowing you to build the version of boot you're using to build i. (E.g. Building 2.5.0 with 2.5.0)

Also there was a problem with aot compilation, I'm not sure if this is related to the version issue above. Maybe @micha remembers...

martinklepsch:
Closing this also so we can use #340 as canonical issue

PR #359, opened 2015-12-16
Convert url only repos to maps in push task

### LINKED ISSUE #358: Push repo-map option doesn't work
Repo-map is being merged with `r` which will be a string in case of default Clojars repo.

Deraen:
Repositories can be defined as maps or as urls. Default Clojars repo is
defined as a url. The string containing the url can't be merged with
`repo-map`. This change adds the same logic to push task as is used in
aether code to convert urls to maps.

Fixes #358


micha:
:beers: :+1:

PR #372, opened 2015-12-26
Checkout task - Remove existing deps

flyboarder:
When adding checkout dependencies, task should remove existing dependencies of the same symbol.


flyboarder:
Superseded by 2.6.0-SNAPSHOT

Issue #375, opened 2015-12-31
Add option to show task to verify dependencies


tobias:
Similar to `lein deps :verify`. 

I have this working locally, I just have a couple of questions about how you would like to see the output. [tcrawley/boot-verify](https://github.com/tobias/boot-verify) just prints a flat list:

![e9276](https://cloud.githubusercontent.com/assets/2631/12064856/35d7ec92-af9b-11e5-89ff-e3c65d7b28d2.png)

My current `boot show --verify-deps` implementation shows the dependency tree with each dep prefixed with the verification status, and is really an enhancement to `boot show -d`:

![d8cc5](https://cloud.githubusercontent.com/assets/2631/12064863/633d442a-af9b-11e5-9413-0aa554028665.png)

Another option is to show the status after each dep:

![eb818](https://cloud.githubusercontent.com/assets/2631/12064893/e025b864-af9b-11e5-8655-e50a4772df68.png)

Yet another option (which I haven't yet tried to implement) is to modify `boot.util/print-tree` to allow for a per-entry prefix, which would allow me to display the verification status on the far left of the tree (which would be a blending of the `boot verify`-style output and `boot show -d`).

Do you have a preference for any of these four output options?

alandipert:
Thinking about this myself, it seems to depend on how we intend people to use the output.

The tree-style output is for human consumption, and so it doesn't matter as much what we put in there.  I'm attracted to "flat" output because it's especially amenable to `awk` etc.

I vote we enhance `boot show -d` as per your second graphic.  It's unlikely anyone is consuming the output programmatically, and I feel personally that it's just as easy to read, especially with the coloring.

Other than that... maybe what we also want is an EDN output for `show` - dump everything we know about dependencies in a machine-readable format.

micha:
I like the second example, but the signing status keywords are different lengths, which interferes with the ability to see the dependency graph clearly. So I think the third one is really the most useful. If you'd like to try putting the signature status on the far left I can probably add something to `print-tree` for that.

micha:
Oh, also :+1: for just adding this functionality to `show -d`.

tobias:
@alandipert - I agree - I think it would be useful to have an option to show that would emit whatever was requested as EDN - I'll file an issue for that.

I went ahead and modified `print-tree` to allow for a prefix fn, just to see what that would look like, and now have:

[...]

PR #377, opened 2015-12-31
Add option to show task to verify deps [#375]

tobias:
This also enhances `boot.util/print-tree` to allow per-node prefixing
and custom node display.

### DIFF
--- a/boot/aether/src/boot/aether.clj
+++ b/boot/aether/src/boot/aether.clj
@@ -148,13 +148,90 @@
   [env]
   (->> env jars-dep-graph ksort/topo-sort reverse))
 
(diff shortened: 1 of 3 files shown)


micha:
Awesome! I wonder if we should just make `verify?` the default behavior for `show -d`? Seems like the extra information wouldn't hurt, right?

Deraen:
Looking at the implementation, I would think it takes some time to retrieve the signature data etc. -> Increased run time hurts a bit.

tobias:
@Deraen is correct - verifying deps is slow, since we have to do an aether resolve for every dependency individually.

PR #380, opened 2016-01-07
Support SSL certificates for repositories

crisptrutski:
This feature is the only blocker to using `boot` for some projects currently, and if it's already supported I couldn't figure out how :smile: 

Ported code from [here]( https://github.com/technomancy/leiningen/blob/d701275464c185f4f22be58450b1af4b94a79836/leiningen-core/src/leiningen/core/ssl.clj)

~~Since you'd want to add these certs before the dependencies are added, makes more sense as an `add-env!` hook than a task, but starting with the simplest step first.~~ 

Switched

~~At first tried to build this as a plugin, but had trouble getting the worker pod to add dependencies.~~ 

Think it makes more sense to add to boot itself, to integrate with `set-env!` (and show up in `show -e`)

### DIFF
--- a/boot/aether/project.clj
+++ b/boot/aether/project.clj
@@ -16,4 +16,6 @@
   :dependencies [[org.clojure/clojure      "1.6.0"  :scope "compile"]
                  [boot/base                ~version :scope "provided"]
                  [boot/pod                 ~version :scope "compile"]
-                 [com.cemerick/pomegranate "0.3.0"  :scope "compile"]])
+                 [com.cemerick/pomegranate "0.3.0"  :scope "compile"]
+                 ;; updated wagon-http to lein's version, so the code could port
+                 [org.apache.maven.wagon/wagon-http "2.9"]])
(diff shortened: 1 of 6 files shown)


crisptrutski:
Think this it good-to-go according to my limited taste. 

Not sure why the builds started failing - looks like travis isn't using a dependency cache and is timing out :disappointed:

boot/aether/project.clj
```diff
@@ -16,4 +16,5 @@
   :dependencies [[org.clojure/clojure      "1.6.0"  :scope "compile"]
                  [boot/base                ~version :scope "provided"]
                  [boot/pod                 ~version :scope "compile"]
-                 [com.cemerick/pomegranate "0.3.0"  :scope "compile"]])
+                 [com.cemerick/pomegranate "0.3.0"  :scope "compile"]
+                 [org.apache.maven.wagon/wagon-http "2.9"]])
```
martinklepsch: Should this be scoped like the other dependencies?

micha: `"compile"` is the default, I don't know why I put those `:scope "compile"` things in there in the first place :)

crisptrutski: This entry is just to bump up the version already included by pomegranite - so sounds like yes, should have the same scope

PR #382, opened 2016-01-11
Mention nix package in readme.

ragnard:
Was discussed briefly on slack as apparently more than one person (me) is using the nix package (yay!).


alandipert:
Great, thanks!

Issue #388, opened 2016-01-16
Keybase support


instilled:
Setting up the GPG toolchain is cumbersome (on OSx even more so). With [keybase.io](https://keybase.io) things got a lot easier recently. That's why I was wondering if there's an interest in bringing `keybase` into `boot`. 

Changes to the existing `boot.pod.gpg` namespace would be marginal. I've already hacked something together that hooks into the existing `boot.core.built-in/push` task. Actually it `alter-var-root`s `boot.pod.gpg/sign-jar` to `instilled.boot.keybase/sign-jar` in a task, see [here](https://github.com/instilled/boot-ext/blob/master/src/main/clojure/instilled/boot/task.clj#L11). 

Instead of that hackery, if there's interest into bringing this into boot, we should find a better way to accomplish this. An env var à la `BOOT_SIGN_TOOL={gpg,keybase}` or an additional parameter to `boot.core.built-int/push` and a minor rework of `boot.pod.gpg` would probably just do fine.

Let me know what you think!

Cheers, Fabio

micha:
@instilled I think this is a great idea!

I have been thinking that the `push` task does way too many things currently. What do you think about separating the jar signing into a separate task and deprecating the GPG stuff that's built into the existing `push` task? @Deraen thoughts?

Deraen:
Yes, separating gpg stuff makes sense.

instilled:
@micha Awesome! Extracting jar-signing into a separate task does make a lot of sense to me. Would you then rather go for two distinct tasks, one for each signing backend (sign-gpg, sign-keybase) or having only one that is configurable?  In my opinion one task should just do fine as the changes really are minimal.

Deraen:
I think one task makes sense. It is possible that a single project will be deployed by several users who have different preferences on how to sign packages.

instilled:
@Deraen @micha  I have pushed the necessary changes to gpg namespace but have not splitted the push task. To do that I would require some guidance, mainly on how to communicate the `articat-map` (see [here](https://github.com/boot-clj/boot/blob/master/boot/core/src/boot/task/built_in.clj#L779)) from the sign to the push (set-env! maybe?).

PR #389, opened 2016-01-16
Adding keybase support #388

instilled:
This is the minimal change required to make signing through keybase work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Clojure projects from scripts written in Clojure that run in the context of
 the project.
 
 > If you have questions or need help, please [visit our Discourse site][discourse].
-> You can find other developers and users in [the `#hoplon` channel on freenode IRC][irc]
+> You can find other developers and users in [the `#bootclj` channel on freenode IRC][irc]
 > and [the `#boot` channel on Clojurians Slack][slack].
 
 ### Another Build Tool?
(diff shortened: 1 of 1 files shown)


Deraen:
I think we should just merge this now and release this in a patch release. We can split the push task in 2.6 release.

Deraen:
@instilled I've been experimenting with Keybase myself and in my set up I have gpg private key locally available using `gpg` so there is no need for me to use `keybase` to sign releases. Is there a set up where private key is not available to `gpg`?

arichiardi:
@Deraen I think `keybase` might be a gpg replacement, meaning that it creates and manages the private keys for you, overriding `gpg`. I am guessing this because when I did my `keybase login` I was asked either to use `gpg` commands (`keybase` effectively wrapping `gpg`) or import my private key in `keybase` (full control in keybase). But of course it would be better to ask around if someone is using a vanilla `keybase` with no `gpg`...

instilled:
@Deraen No, not really. It will probably take quite a while until keybase support signing without shelling out to GPG. This will require importing keys into the local GPG registry anyways. My motivation for using keybase is the easy setup and the command line interface that feels more natural to me. Also it will ask for the key's pass with copy&paste support if GPGTools are available (I manage my passwords in vault) as somehow with the native GPG I fail to paste it in the prompt. Otherwise than that there's no other benefit in using keybase instead of native GPG.

PR #398, opened 2016-01-26
filter files by metadata/properties

dm3:
A complement to the `by-name/by-re/by-ext` filters. 

Pretty self-explanatory. However, `by-meta` differs from the existing filters - it doesn't unwrap the `TmpFiles` into `java.io.File` objects and doesn't work on `java.io.File`s as they have no metadata.


arichiardi:
:+1:

PR #400, opened 2016-01-27
Consider unpeeled Git tag references

malesch:


micha:
Thanks, I _think_ I understand...I always create tags via `git tag -a ...`, which creates annotated tag objects. But you could just use `git tag ...` (without the `-a` option) which creates a reference that needs to be peeled, because it could point to another tag, for example?

malesch:
Exactly. I am not a Git expert but this is also my understanding. With `-a` you create an _annotated tag object_ in contrast to so a called _lightweight tag_, which is only a simple ref (created without one of the options `-a`, `-s` or `-m`).

martinklepsch:
Looks good to me :+1:

Issue #291, opened 2015-09-15
boot uber fails if dependency tree contains POM artifacts


wagjo:
Minimal example:
`boot -d org.apache.jena/jena-text:3.0.0 -- pom -p my-project -v 0.1.0 -- uber`

fails with `clojure.lang.ExceptionInfo: java.util.zip.ZipException: error in opening zip file`

jena-text depends on `org.apache.jena/apache-jena-libs:3.0.0` which is a POM artifact - it does not contain any .jar per se. `uber` task should correctly handle such dependencies.

micha:
I believe this has been fixed with #292. Please reopen if this is not the case. Thanks!

PR #403, opened 2016-02-04
Do not include POM artifacts in uberjar

wagjo:
Implementation of uber task has been changed and #291 bug reemerged. After following patch POM artifacts are removed from the list of jars that are to be included in uberjar.


micha:
Thanks @wagjo, sorry about the regression!

PR #406, opened 2016-02-05
pom packaging type & aether extension

instilled:
This enables pom to provide packaging type other than the default (jar) and modifies `aether/{deploy,install}` fns by adding :extension to jarpath in artifact-map based on either packaging type in pom or file extension.

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -321,6 +321,7 @@
   [p project SYM           sym         "The project id (eg. foo/bar)."
    v version VER           str         "The project version."
    d description DESC      str         "The project description."
+   P packaging STR         str         "The project packaging type, i.e. war, pom"
    u url URL               str         "The project homepage url."
    s scm KEY=VAL           {kw str}    "The project scm map (KEY is one of url, tag, connection, developerConnection)."
    l license NAME:URL      {str str}   "The map {name url} of project licenses."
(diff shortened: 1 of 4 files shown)


alandipert:
Thanks for this!  Can you elaborate a little more on the reason one might want to use pom together with anything but jar?  I'm just trying to understand the use case.  Thanks in advance for clarifying.

instilled:
Jup. Boot being an awesome build tool I think it should also support publishing other types of artifacts. With the status quo boot will never push anything but `jar`s to a maven repository. Personally I would love to see boot also supporting at least `war`s and maybe other binaries such as `zip` or `tar.gz` files (thinking of assemblies here). By supporting multiple filetypes we may also want to add support for classifiers which allow multiple artifacts of the same version to be pushed to the maven repository (see [here](https://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html) or [here](http://stackoverflow.com/a/20909695) for further details on classifiers).

boot/core/src/boot/task/built_in.clj
```diff
@@ -331,7 +333,12 @@
         tag  (or (:tag scm) (util/guard (git/last-commit)))
         scm  (when scm (assoc scm :tag tag))
         deps (or dependencies (:dependencies (core/get-env)))
-        opts (assoc *opts* :scm scm :dependencies deps :developers developers)]
+        opts (assoc *opts*
+               :scm scm
+               :dependencies deps
+               :developers developers
+               :classifies classifier
```
micha: Is this a typo? Should `:classifies` here be `:classifier`?

PR #407, opened 2016-02-05
Unpeel ExecutionExceptions

arichiardi:
After discussion on Slack it emerged that it is necessary and robust to unpeel ExecutionExceptions
when dereffing a feature so that the App.Exit (or whatever) inside can be properly propagated.
   
After tracking down all the @ and deref, I came to the conclusion that only boot.core/boot, the
entry point of the execution, needs this.
    
This patch does exactly that.

(hopefully correctly)


Deraen:
Merged: https://github.com/boot-clj/boot/commit/93c00ed92ce8624c421cf691b24a56db94ebf86f

I rebased the commit to keep old indentation.

PR #413, opened 2016-02-14
target task: default output to :target-path if defined, otherwise "target"

mobileink:
"target" is currently hardcoded in the target task.  This fixes that.  I don't really know how to do a proper pull request so let me know if I need to do something different.


micha:
@mobileink thanks for the PR!

However, I'm not sure if this is needed, since we are deprecating `:target-path` and the implicit target behavior. I think the migration should be to use the following instead of `:target-path` in the env:

```clojure
(task-options!
  target {:dir #{"target"}})
```

What do you think?

alandipert:
We'll be deprecating implicit target soon, so I also think we don't quite need this.  Please don't hesitate to reopen if you feel differently.

PR #414, opened 2016-02-15
task aot: add option -d compile-path-sfx

mobileink:
This option allows the user to set *compile-path*, which is
where compiled .class files go.  The specified path will be
relativized if necessary and suffixed to the build root dir.

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -549,9 +549,13 @@
   "Perform AOT compilation of Clojure namespaces."
 
   [a all          bool   "Compile all namespaces."
-   n namespace NS #{sym} "The set of namespaces to compile."]
+   n namespace NS #{sym} "The set of namespaces to compile."
+   d compile-path-sfx PATH str "*compile-path* suffix"]
 
   (let [tgt         (core/tmp-dir!)
+        sfx (if compile-path-sfx (clojure.string/replace compile-path-sfx #"^/" "") nil)
+        sfx (if (empty? sfx) nil sfx) ;; handle pathological case -d "/"
+        compile-path (if sfx (str (.getPath tgt) "/" sfx) (.getPath tgt))
         pod-env     (update-in (core/get-env) [:directories] conj (.getPath tgt))
         compile-pod (future (pod/make-pod pod-env))]
     (core/with-pre-wrap [fs]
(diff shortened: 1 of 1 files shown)


boot/core/src/boot/task/built_in.clj
```diff
@@ -549,17 +549,23 @@
...
+        compile-path (if sfx (str (.getPath tgt) "/" sfx) (.getPath tgt))
         pod-env     (update-in (core/get-env) [:directories] conj (.getPath tgt))
         compile-pod (future (pod/make-pod pod-env))]
     (core/with-pre-wrap [fs]
       (core/empty-dir! tgt)
       (let [all-nses (->> fs core/fileset-namespaces)
             nses     (->> all-nses (set/intersection (if all all-nses namespace)) sort)]
         (pod/with-eval-in @compile-pod
-          (binding [*compile-path* ~(.getPath tgt)]
+          ~(io/make-parents (io/file tgt (if sfx (str sfx "/dummy") "")))
```
micha:
```clojure
~(.mkdirs (io/file tgt (str sfx)))
```

micha:
Do we need to relativize the path? Seems ok if we have an assertion instead, to avoid confusion.

Also it seems more like a _prefix_ for the generated class files than a _suffix_ for the `*compile-path*`, but maybe my brain is backwards :). Perhaps something like `target-dir` or something might be a good name for the option?

Could be do the subdirectory earlier, like this?

```clojure
(core/deftask aot
  "Perform AOT compilation of Clojure namespaces."
  [a all          bool   "Compile all namespaces."
   n namespace NS #{sym} "The set of namespaces to compile."
   d compile-path-sfx PATH str "*compile-path* suffix"]
  (let [tmp         (core/tmp-dir!)
        tgt         (io/file tmp (str compile-path-sfx))
        pod-env     (update-in (core/get-env) [:directories] conj (.getPath tgt))
        compile-pod (future (pod/make-pod pod-env))]
    (core/with-pre-wrap [fs]
      (core/empty-dir! tgt)
      (let [all-nses (->> fs core/fileset-namespaces)
            nses     (->> all-nses (set/intersection (if all all-nses namespace)) sort)]
        (pod/with-eval-in @compile-pod
          (binding [*compile-path* ~(.getPath tgt)]
            (doseq [[idx ns] (map-indexed vector '~nses)]
              (boot.util/info "Compiling %s/%s %s...\n" (inc idx) (count '~nses) ns)
              (compile ns)))))
      (-> fs (core/add-resource tmp) core/commit!))))
```

micha:
What is the use case for this? It will cause problems with the classpath wouldn't it? Like what happens when you add the class files with "false paths" to the classpath of the main boot pod after compiling?

micha:
This seems like more of a job for the sift task, if anything.

PR #428, opened 2016-03-10
Integrate boot-new into Boot core

seancorfield:
This adds the `new` task to `task/built-ins` and adds all the files needed to support `boot new` templates (the four built-in templates: `app`, `default`, `task`, `template`) as well as `new_helpers.clj` with supporting logic for `boot new` and the `leiningen.new.templates` and `boot.new.templates` namespaces expected by templates for project generation.


seancorfield:
This fails because the build tries to AOT compile everything and there are references to Stencil in namespaces that are (deliberately) pulled in dynamically so the dependency on Stencil can be deferred to runtime (and hence only when `boot new` is actually run).

I'll see if I can remove that require and allow it to be AOT-compiled.

seancorfield:
Duh! Loading the namespace still runs the `def` which causes Stencil to be required... I'll have a think about how to support AOT compilation but still only bring in Stencil at runtime.

alandipert:
We discussed in Slack and iirc, came to consensus that it's a good idea to hold off on this for now and revisit in the future.

seancorfield:
:+1:

PR #434, opened 2016-03-31
Check that each keyword arg has a supplied value

elogent:
I'm just starting to learn Boot, and I spent a few hours scratching my head when `(aot :all)` or `(show :fileset)` gave different results from `(aot "-a")` or `(show "-f")`. This patch throws an exception if the last keyword argument to a task is not supplied a value.

### DIFF
--- a/boot/core/src/boot/cli.clj
+++ b/boot/core/src/boot/cli.clj
@@ -167,6 +167,8 @@
     (cond
       (nil? arg) split
       (not (keyword? arg)) (recur (update-in split [:cli] (fnil conj []) arg) more)
+      (empty? more) (update-in split [:errors] (fnil conj [])
+                               (str "no value supplied for option " arg))
       :else (recur (update-in split [:kw] assoc arg (first more)) (rest more)))))
 
 (defmacro ^:private assert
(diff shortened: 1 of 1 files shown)


Deraen:
Looks good to me :+1:

micha:
There is still a problem when there are two flags, isn't there?

Example:

```
boot foo -x -y
```

```clojure
(boot (foo :x :y))
```

elogent:
That case is handled by the type checker. For instance, if I do this:

```clojure
(deftask foo
  [x xenon bool "Add xenon gas."
   y yttrium bool "Add yttrium metal."]
  (prn *opts*))

(deftask bar []
  (foo :xenon :yttrium))
```

I'll get this error:

```
$ boot bar
        clojure.lang.ExceptionInfo: option :xenon must be of type bool
    data: {:file "/tmp/boot.user4034053871589354536.clj", :line 11}
java.lang.IllegalArgumentException: option :xenon must be of type bool
                boot.user/eval65/fn  boot.user4034053871589354536.clj:    5
                                ...                                        
                boot.user/eval94/fn  boot.user4034053871589354536.clj:    7
                                ...                                        
    clojure.core/apply/invokeStatic                          core.clj:  646
                 clojure.core/apply                          core.clj:  641
          boot.core/construct-tasks                          core.clj:  891
                                ...                                        
    clojure.core/apply/invokeStatic                          core.clj:  646
                 clojure.core/apply                          core.clj:  641
                  boot.core/boot/fn                          core.clj:  936
clojure.core/binding-conveyor-fn/fn                          core.clj: 1938
                                ...
```

micha:
Oh, right! Excellent! Thanks for clarifying :)

PR #436, opened 2016-04-05
Add boot.util/map-as-dep

arichiardi:
The complement of boot.util/dep-as-map, takes a dependency map and returns a dependency vector.


micha:
:+1: looks good to me!

arichiardi:
:+1:

PR #437, opened 2016-04-06
Parallel boot

arichiardi:
Adds both parallel computation of tasks and parallel testing to boot!

Works smoothly but it is still a bit verbose at the moment, cause as discussed on Slack it is maybe better not to call `boot.main/-main` every time and therefore modify `runboot` accordingly...will take care of that.


micha:
@arichiardi Is this ready to merge?

arichiardi:
The basic functionality is these but it still calls `boot.main/-main` in `runBoot `

PR #463, opened 2016-05-16
Fix boot.test/runtests, ns-interns is evaluated at task execution

arichiardi:
In order for the autodiscovery of deftesttask to work, an ns-interns call is necessary. The problem
was that before this patch boot.test/runtest was executing it at task creation time, necessitating a
require in build.boot for it to work.

Potentially though, the require of the test can be done in a separate task right before calling
boot.test/runtest.

This is why ns-interns has been moved at task execution time (inside a with-pass-thru, storing in an
atom the detected commands).


micha:
:100: sweet!

PR #476, opened 2016-07-02
Do not force Vicki’s voice for notifications

skammer:
Vicki was a nice voice a while back, but now that Apple finally supplied us with wide variety of high quality voices, it seems such a waste to torture us with Vicki's robotic pronunciation.

By using systems default voice user themselves can decide which voice he likes best.

Or maybe we can have a setting for this somewhere, which seems to be an overkill for such simple task.


micha:
Makes sense to me :+1:

PR #480, opened 2016-07-13
Print stack trace for "unexpected" exception

seancorfield:
If Boot throws a non-Exit exception, print a stack trace to help user
debug the problem. This happens with older Clojure versions for example
but may also happen with damaged Boot installations perhaps?


micha:
Awesome, thanks!

PR #488, opened 2016-08-03
Fix path format for `fake.class.path` on Windows

oahner:
The `fake.class.path` env variable is currently populated in a way that
leaves an extraneous leading forward slash character at the start of all
path components on Windows, so that:

  "C:\\Program Files\\Java\\jdk1.8.0_77\\jre\\lib\\ext\\access-bridge-64.jar"

Ends up being represented as:

  "/C:/Program Files/Java/jdk1.8.0_77/jre/lib/ext/access-bridge-64.jar"

This patch uses `java.nio.file.Paths` to properly convert each component
of the classpath to a valid system-specific path before using it to
populate the `fake.class.path` variable.

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -21,7 +21,7 @@
   (:import
     [boot App]
     [java.io File]
-    [java.nio.file Path]
+    [java.nio.file Path Paths]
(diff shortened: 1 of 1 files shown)


micha:
Looks great, thanks! :+1:

PR #490, opened 2016-08-08
Add string type hint for Exception constructor in deftask

stephenbrady:
Fixes reflection warnings


micha:
Good catch! Thanks! :+1:

PR #498, opened 2016-08-28
extend dep. mgt. functions; add unit tests

loomis:
Provides unit tests of the `dep-as-map` and `map-as-dep` utility functions.  Verifies that the fix in the accepted PR #479 works correctly.

In addition, this PR extends these functions to treat dependency vectors that do not have a version specified, such as:
```
['example/project]
['example/project :scope "test"]
```
Such incomplete specifications are useful when handling dependency versions and options centrally. At SixSq, we use a function to pull in a central list of dependency specifications to complete any dependencies in the local `build.boot` file.

The change also will allow options with nil values to be retained.  This may not be useful, but it eliminates a case where one function could produce a value that wasn't accepted by the other.  Previously,
```
(map-as-dep {:project 'example/project :version "1.0.0" :option nil})
;; produced
;; ['example/project "1.0.0" :option]
```
which wasn't valid.  It now produces:
```
['example/project "1.0.0" :option nil]
```

### DIFF
--- a/boot/pod/src/boot/util.clj
+++ b/boot/pod/src/boot/util.clj
@@ -327,20 +327,22 @@
 
 (defn dep-as-map
   "Returns the given dependency vector as a map with :project and :version
-  keys plus any modifiers (eg. :scope, :exclusions, etc)."
-  [[project version & kvs]]
-  (let [d {:project project :version version}]
-    (merge {:scope "compile"}
-      (if-not (seq kvs) d (apply assoc d kvs)))))
+  keys plus any modifiers (eg. :scope, :exclusions, etc).  If the version
+  is not specified, nil will be used."
+  [[project & terms]]
+  (let [[version & kvs] (if (odd? (count terms)) terms (cons nil terms))
+        d {:project project :version version :scope "compile"}]
+    (if-not (seq kvs) d (apply assoc d kvs))))
 
 (defn map-as-dep
   "Returns the given dependency vector with :project and :version put at
-  index 0 and 1 respectively and modifiers (eg. :scope, :exclusions,
-  etc) next."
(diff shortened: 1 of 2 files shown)


loomis:
@micha I've not looked at the new additions to Pomegranate, so I can't judge their usefulness.

Our approach has been to keep a "dependency management" list in a separate artifact (just a normal vector of dependency specifications).  This artifact is then pulled in and merged with the dependencies in the `build.boot` file.  Our dependency declarations end up looking like:
```
(set-env!
  :dependencies
  #(vec (concat %
                (merge-defaults
                 ['sixsq/default-deps (get-env :version)]
                 '[[org.clojure/clojure]

                   [aleph]
                   [cheshire] ;; newer version needed for ring-json
                   [compojure]
                   [clj-stacktrace]
                   [tolitius/boot-check]]))))
```
where the `merge-defaults` function is defined [here](https://github.com/SixSq/boot-tasks/blob/master/build-utils/src/sixsq/build_fns.clj#L22).

I think that coming up with a common approach for managing dependencies across modules in boot would be useful.  If the updates to Pomegranate can help, that's great.  What's the best way to come to a consensus for a solution?

flyboarder:
@loomis @micha I agree it would be nice to have a few more utils for deps management, currently I place this in every boot project I have, would be nice if there was a proper way to do this without cluttering my boot file:

```clojure
(defn read-file   [file] (read-string (slurp file)))
(defn get-deps    []     (read-file "./dependencies.edn"))
(defn get-devdeps []     (read-file "./dev_dependencies.edn"))

```

loomis:
@micha @flyboarder I've submitted a PR (#507) with an implementation of managed dependencies for boot. This also allows dependencies to be read from a resource.  Feedback welcome in the other PR.

PR #507, opened 2016-09-17
implement managed dependencies

loomis:
This PR allows the use of "managed dependencies" when defining the dependencies for a project.  This is a useful feature when trying to maintain consistent dependency versions, scopes, exclusions, etc. between a number of different projects.

Although most people maintain the dependency list directly within the `build.boot` file itself, this PR will also allow dependencies to be read in from an external file, either in the project hierarchy or from a maven artifact. The typical usage in this case is:

```
(require '[boot.deps :as deps])

(set-env!
  :dependencies
    (deps/load "deps"))
```
This will load the `dependencies.edn` file from the `deps` subdirectory within the project.  The `deps` subdirectory and the `dependencies.edn` file will not be added to the project's fileset. 

This will probably be more often used to read in a list of managed dependencies.  In this case, the usage would be:

```
(require '[boot.deps :as deps])

(set-env!
  :managed-dependencies
    (deps/load 
      '[example/deps "1.0.0"]
      :resource "managed-dependencies.edn")
  :dependencies 
    '[[foo/bar]
      [baz]])
```
Here, the name of the resource has been changed and the file with the dependencies is instead loaded from the defined artifact.  Again, the dependency file and the artifact are not added to the project.  In this case, the versions and perhaps other options will be merged into the (incomplete) dependencies listed in `:dependencies`. 

It is also possible to transform the loaded dependencies before passing them on to `boot`.  For example,

```
(require 
   '[boot.deps :as deps]
   '[clojure.walk :as walk])

(set-env!
  :managed-dependencies 
    (deps/load 
      '[example/deps "1.0.0"]
      :resource "managed-dependencies.edn"
      :xfn #(walk/postwalk-replace {:version "1.2.3"} %))
  :dependencies
    '[[foo/bar]
      [baz]])
```
In this case, any`:version` keywords will be replaced by "1.2.3".  This is often useful for cross-project dependencies where the projects share a common version number.  Rather than a hard-coded value, this would probably be taken from the project version definition. 

Some comments about the implementation:

 * This includes an update to pomegranate v0.3.1.  This version implements managed dependencies. Unfortunately, it also allows versions to be managed (not other characteristics, especially exclusions). It also doesn't appear to allow jar and non-jar dependencies to be managed separately.  These are the reasons why this PR implements the managed dependencies itself, rather than relying on the pomegranate implementation. 

 * It updates the `.gitignore` file to exclude files from IntelliJ.  This could be removed from the PR if it isn't desired, but this seems to be a harmless addition (and is useful for me).

 * I've updated the `map-as-dep` and `dep-as-map` functions to also treat correctly the "extension" value for the dependency. 

 * There were a number of duplicate utilities in the `boot.pod` and `boot.util` namespaces.  I needed the `canonical-id` utility in `boot.util`, so I've moved all of the related functions there.  I've created aliases in the `boot.pod` namespace for backwards compatibility.  I haven't marked these as deprecated, but that would be a possibility.  Note that I've also mapped `coord->map` and `map->coord` to `util/dep-as-map` and `util/map-as-dep`, which provide more correct implementations. 

 * Tests have been provided for nearly all of the new functions.  I also had to correct tests in `boot.pod-test` related to the constructors for filesets. The fileset tests pass now, but I haven't verified that they do something reasonable. The tests can be run by hand, but don't appear to run automatically from the Makefile build.

 * One detail that I don't like is that I had to deal with an added path differently from an added dependency when creating the pod to isolate these changes.  Perhaps this is a misunderstanding on my part, but I expected just adding a new source path to the environment before creating the pod would update the pod's classpath.  This didn't work, do I've updated the classpath explicitly in the `read-deps` function. 

I'd be interested in feedback on the feature and its implementation.  This works well for my use case, but I'd like to understand if there's something I can change to ensure it is as widely useful as possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,9 @@ pom.xml.asc
 /zzz/
 /launch4j-config.xml
 /launch4j/
+
+### INTELLIJ ###################################################################
+
+.idea
+*.iml
+
--- a/boot/aether/project.clj
+++ b/boot/aether/project.clj
@@ -16,5 +16,5 @@
   :dependencies [[org.clojure/clojure               "1.6.0"  :scope "compile"]
                  [boot/base                         ~version :scope "provided"]
                  [boot/pod                          ~version :scope "compile"]
-                 [com.cemerick/pomegranate          "0.3.0"  :scope "compile"]
+                 [com.cemerick/pomegranate          "0.3.1"  :scope "compile"]
                  [org.apache.maven.wagon/wagon-http "2.9"    :scope "compile"]])
--- a/boot/aether/src/boot/aether.clj
+++ b/boot/aether/src/boot/aether.clj
@@ -5,6 +5,7 @@
     [clojure.pprint              :as pprint]
     [cemerick.pomegranate.aether :as aether]
     [boot.util                   :as util]
+    [boot.deps-util              :as deps-util]
     [boot.pod                    :as pod]
     [boot.gpg                    :as gpg]
     [boot.from.io.aviso.ansi     :as ansi]
@@ -104,7 +105,7 @@
   [env]
   (try
     (aether/resolve-dependencies
-      :coordinates       (:dependencies env)
+      :coordinates       (deps-util/merge-deps (:dependencies env) (:managed-dependencies env))
       :repositories      (->> (or (seq (:repositories env)) @default-repositories)
                            (map (juxt first (fn [[x y]] (if (map? y) y {:url y}))))
                            (map (juxt first (fn [[x y]] (update-in y [:update] #(or % @update?))))))
--- a/boot/core/project.clj
+++ b/boot/core/project.clj
@@ -16,4 +16,3 @@
   :dependencies   [[org.clojure/clojure "1.6.0"  :scope "provided"]
                    [boot/base           ~version :scope "provided"]
                    [boot/pod            ~version :scope "compile"]])
-
--- a/boot/core/src/boot/deps.clj
+++ b/boot/core/src/boot/deps.clj
@@ -1,0 +1,26 @@
+(ns boot.deps
+  (:refer-clojure :exclude [load])
+  (:require
+    [boot.core :as boot]
+    [boot.pod :as pod]))
+
+(defn load
+  "Loads dependency specifications from a resource inside the given
+   dependency or on the given path.  Keyword options can be provided
+   after the dep-or-path.  The supported options are:
+     :resource The resource name to use to read in the dependencies.
+               If the resource is not provided, then the default
+               \"dependencies.edn\" will be used.
+     :xfn      A function used to transform the loaded dependencies.
+               Takes a list of dependencies and returns a list of
+               transformed dependencies.  Defaults to the identity
+               function."
+  [dep-or-path & {:keys [resource xfn]
+                  :or   {xfn identity}}]
+  (let [path (when (string? dep-or-path) dep-or-path)
+        dep (if (vector? dep-or-path) [dep-or-path] [])]
+    (-> (boot/get-env)
+        (assoc :dependencies dep)
+        pod/make-pod
+        (pod/with-call-in (boot.deps-util/read-deps ~path ~resource))
+        xfn)))
--- a/boot/pod/project.clj
+++ b/boot/pod/project.clj
@@ -18,4 +18,5 @@
   :dependencies   [[boot/base                               ~version :scope "provided"]
                    [org.clojure/clojure                     "1.6.0"  :scope "provided"]
                    [org.tcrawley/dynapath                   "0.2.3"  :scope "compile"]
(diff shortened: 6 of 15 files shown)


flyboarder:
I only took a quick breeze through the changes but I really like this addition myself! I would like to see the ability to hot reload the dependencies from a jar or deps file, similar to how checkout works. i.e. If I'm working on the main deps package my running projects would also update.

alandipert:
Hi, sorry for taking so long to provide feedback.  I appreciate the length you go to describing your motivation and implementation.

I too feel the need for something like this, but I'm not entirely sure Boot itself is the place for it to live.  It seems like you concluded Boot was the place because Boot (pomegranate) is where Maven machinery happens, and managed dependencies are historically Maven machinery.  But couldn't a system with the same affordances of Maven's managed dependencies be crafted and delivered as a separate library?

I ask because I saw https://github.com/borkdude/boot-bundle recently, which seems to offer a nice convention for creating and sharing managed dependency sets as a library.

I don't mean to discourage you by questioning whether or not this should go into Boot.  It's just that we have used Boot every day for years, and hope to use it for many more, and so have to be very sure about feature PRs that come in.  We'd hate to wake up to a Boot we can no longer understand or maintain.  That's why I think it's generally best to try out things like this separately, as libraries, before going into Boot, when possible.

flyboarder:
@loomis is there a way some of this can be ported to a separate boot task?

alandipert:
And even if it can't, I'm still curious to know the advantage of using pomegranate for managed deps over the boot-bundle approach.

flyboarder:
@alandipert the versions in managed dependencies are used when resolving transitive dependencies by pomegranate

where boot-bundle is creating a dependency list by merging multiple lists

alandipert:
Oh, so I was wrong, really this feature is mostly unrelated to boot bundle.  The essential feature here is the ability to influence the versions of transitive deps, which seems only possible by interacting directly with pomegranate.

Seems pretty awesome now that I understand.  @micha what do you think?

loomis:
@alandipert @flyboarder There are three somewhat independent parts of this PR: a "read-deps" utility, "fixes" for boot's dependency-related utilities, and the actual managed dependency feature.  Let me give my opinions on these in reverse order.

Dependency Management
--------------------------

The rationale for trying to use the new pomegranate managed dependency machinery was to 1) take advantage of work already done, 2) be consistent with the low-level mechanisms used by maven (and now leiningen), and 3) to provide full dependency options management (e.g. exclusions in addition to dependency versions).

All of this can, of course, be done via a separate library.  In fact, we already have an internal library of boot functions/tasks that we use for this.  This PR was a generalization and simplification of that library for wider use.  The direct integration with boot, rather than using an external library, simplifies the dependency management by reducing the boilerplate needed in the build file 1) to access the external library and 2) to directly manipulate the dependency data structures to shoehorn them into the :dependencies field.

Unfortunately, the managed dependency feature in pomegranate is rather limited in that it only manages dependency versions and ignores all of the other dependency options.  This is insufficient for our use case, which was the reason for doing more work in the proposed code and the more complicated PR. BTW, neither the PR's implementation nor the pomegranate implementation will affect transitive dependencies (unfortunately), because only the resolved explicit dependencies (not the managed dependency list) are passed down to the lowest levels of the dependency resolution code.

Although boot-bundle is interesting, it focuses on reusing bundles of dependencies rather than ensuring consistency of dependency options between different projects.  The developer has just added a way to provide some management of versions, but the way of doing this is going to be very messy (IMO) if it is extended to scopes, exclusions, optional flags, etc. IMO dependency options management and bundle management are not the same, so these features are more complementary than competitive. 

I still think adding dependency management to boot is worthwhile.  I would suggest doing the upgrade to the latest pomegranate release and make the two line change to use its dependency management.  I'm happy to create a new PR for this, but would like to know if this is worthwhile before doing the work.  The downside is that this will provide only version management, not full options management.  I can approach the pomegranate developers to see if they'd welcome enhancing their code to support full options management. If they would, then boot would benefit from any changes without changing boot itself.

Dependency Related Utilities
-----------------------------

In making the PR, the dependency-related boot utilities needed to be enhanced.  If the dependency management is done in a separate library or in pomegranate, none of these changes are necessary.  Nonetheless, having multiple incompatible implementations in different namespaces (`coord->map`, `map->coord`, `dep-as-map`, `map-as-dep`) is probably worth correcting anyway.  There was also a buggy `canonical-id` function, but I see that there's a separate PR for fixing that. The `full-id` function was also buggy, but perhaps that's fixed in the same PR.

"read-deps" Utility
-------------------

The need to read an edn data structure from a file/resource without adding the file/resource to the project's dependencies is something that came up in this PR, is also reimplemented in boot-bundle, and was an explicit request from @flyboarder initially.  It seems to be common enough, that everyone would benefit from a standard, supported version from boot itself.

PR #514, opened 2016-10-12
Add line numbers to -b option

arichiardi:
I was taking a stab at adding the line number in `-b` for debugging purposes, it looked like a nice and easy feature to add, but I noticed that, while debugging an error I have, sometimes the error reports the line as in "one line per form" and sometimes it just reports the last line number.

Example: `data: {:file "/tmp/boot.user3358451832744193056.clj", :line 97}` where `boot -b|wc -l` is `97`.

So maybe I need a bit more digging and guidance on this one. Also, maybe you guys already tried and abandoned the idea?

### DIFF
--- a/boot/core/src/boot/main.clj
+++ b/boot/core/src/boot/main.clj
@@ -189,7 +189,10 @@
               import-ns   (export-task-namespaces initial-env)
               scriptforms (emit boot? args userforms localforms bootforms import-ns (:init opts))
               scriptstr   (binding [*print-meta* true]
-                            (str (string/join "\n\n" (map pr-boot-form scriptforms)) "\n"))]
+                            (str (->> scriptforms
+                                     (map pr-boot-form)
+                                     (map-indexed (fn [i form] (format "%s %s" i form)))
+                                     (string/join "\n\n")) "\n"))]
 
           (when (:boot-script opts) (util/exit-ok (print scriptstr)))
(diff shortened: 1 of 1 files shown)


micha:
Usually i will do

```bash
boot -b foop |cat -n
```

Also, if you do for example

```bash
boot -v foop
```

and you get an exception with a line number in the build.boot, you should add `-v` to the `-b` call too:

```bash
boot -vb foop |cat -n
```

arichiardi:
On October 12, 2016 11:55:42 AM PDT, Micha Niskin <notifications@github.com> wrote:
>Usually i will do
>
>```bash
>boot -b foop |cat -n
>```
>
>Also, if you do for example
>
>```bash
>boot -v foop
>```
>
>and you get an exception with a line number in the build.boot, you
>should add `-v` to the `-b` call too:
>
>```bash
>boot -vb foop |cat -n
>```
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/boot-clj/boot/pull/514#issuecomment-253305000

Uhm interesting, so when it is the last line...what can it be? (Side question)

Would it be ok to add then the line number directly in `-b`...or maybe add a separate option ?
-ar

arichiardi:
So you're saying it is not worth adding

PR #519, opened 2016-10-28
Update boot.from.io.aviso to use 0.1.30

hlship:
This updates the boot.from.io.aviso namespace to use fresh copies of io.aviso/pretty 0.1.30.   

For want of a real process, I used rsync and some hand editing to accomplish this.
Also, the way that frame filters work changed slightly, affecting the boot.utils namespace.

`make clean test` works, so that gives me some confidence.


micha:
@hlship Sorry, I forgot to thank you for this PR, it looks like it was a good amount of work. Thanks!

hlship:
Not a problem, maybe an hour or so to find my way around.

PR #523, opened 2016-11-12
Prevent play! from throwing on systems with no audio output

### LINKED ISSUE #520: Check for available AudioDevice in speak task
I would be great that in `(speak)` task, we use try catch or check for available AudioDevice before running the task to avoid error while using in the environment without AudioDevice.

```
  clojure.lang.ExceptionInfo: javazoom.jl.decoder.JavaLayerException: Cannot create AudioDevice
    data: {:file "/tmp/boot.user3476942433437848616.clj", :line 21}
java.util.concurrent.ExecutionException: javazoom.jl.decoder.JavaLayerException: Cannot create AudioDevice
 javazoom.jl.decoder.JavaLayerException: Cannot create AudioDevice
 javazoom.jl.decoder.JavaLayerException: cannot obtain source audio line
     java.lang.IllegalArgumentException: No line matching interface SourceDataLine supporting format PCM_SIGNED 22050.0 Hz, 16 bit, mono, 2 bytes/frame, little-endian is supported.
                         javax.sound.sampled.AudioSystem.getLine                  AudioSystem.java:  479
            javazoom.jl.player.JavaSoundAudioDevice.createSource         JavaSoundAudioDevice.java:  101
               javazoom.jl.player.JavaSoundAudioDevice.writeImpl         JavaSoundAudioDevice.java:  148
                        javazoom.jl.player.AudioDeviceBase.write              AudioDeviceBase.java:  133
                    javazoom.jl.player.JavaSoundAudioDevice.test         JavaSoundAudioDevice.java:  205
  javazoom.jl.player.JavaSoundAudioDeviceFactory.testAudioDevice  JavaSoundAudioDeviceFactory.java:   83
javazoom.jl.player.JavaSoundAudioDeviceFactory.createAudioDevice  JavaSoundAudioDeviceFactory.java:   42
            javazoom.jl.player.FactoryRegistry.createAudioDevice              FactoryRegistry.java:   87
                                javazoom.jl.player.Player.<init>                       Player.java:   97
                                javazoom.jl.player.Player.<init>                       Player.java:   82
                                            boot.notify/play!/fn                        notify.clj:   23
                             clojure.core/binding-conveyor-fn/fn                          core.clj: 1938
```

RadicalZephyr:
I chose to catch `JavaLayerException` because that constrains the types of
errors that will be caught to those originating inside of Java Layer.

It's possible that we could detect the presence/absence of an audio
device using the Java Sound API AudioSystem, but it's not clear to me
that would be more robust.  This will probably also catch problems like
trying to play a format that is not supported by Java Layer.

Closes #520.

### DIFF
--- a/boot/worker/src/boot/notify.clj
+++ b/boot/worker/src/boot/notify.clj
@@ -17,13 +17,17 @@
 
 (defn play! [file]
   (fg-first-time!
-    (future
-      (-> (or (.getResourceAsStream (clojure.lang.RT/baseLoader) file)
-            (FileInputStream. (io/file file))
-            (or (throw (RuntimeException. (str file " not found.")))))
-        java.io.BufferedInputStream.
-        javazoom.jl.player.Player.
-        .play))))
+   (future
+     (try
+       (-> (or (.getResourceAsStream (clojure.lang.RT/baseLoader) file)
+               (FileInputStream. (io/file file))
+               (throw (RuntimeException. (str file " not found."))))
+           java.io.BufferedInputStream.
+           javazoom.jl.player.Player.
+           .play)
+       (catch Exception e
+         (util/warn "\nError attempting to play sound file: %s\n\n"
+                    (.getMessage e)))))))
 
 (defn success! [theme file] (play! (or file (path-for theme "success"))))
 (defn failure! [theme file] (play! (or file (path-for theme "failure"))))


micha:
@RadicalZephyr Awesome, thanks for the patch!

One question, perhaps it would be better to just catch any exception there, rather than only the specific javazoom one? Seems like there really isn't any exception there that we would want to propagate up the stack any further.

RadicalZephyr:
Yeah, that's a good question.  One reason I caught the JavaZoom one specifically was that there's an explicit throw if the file isn't found.  But on thinking about it today, that's probably only worthy of a `warning` message as well.

I'll change it to catch `Exception`

RadicalZephyr:
@micha FYI, I updated this to catch `Exception` now.

PR #526, opened 2016-11-20
use pomegranate for managed dependencies

loomis:
Allow boot users to specify a set of managed dependencies (under the `:managed-dependencies` key) as well as the existing `:dependencies`.  Pomegranate uses the managed dependency information to complete any missing versions in the dependencies and may be used to influence the versions for transient dependencies.  This feature is useful when trying to maintain a set of consistent dependency versions across a number of different projects.


Deraen:
Looks simple enough.

Leiningen also supports this since 2.7.0.

PR #531, opened 2016-11-25
Improve io.aviso/pretty support

hlship:
Updates the repackaged io.aviso/pretty to 0.1.33.
 
Changes how nrepl works:
- Pretty exception output is always enabled
- If colorization is disabled, then ANSI font colors are disabled (but the rest of pretty is still in effect)

Note: wasn't able to get BOOT_COLOR to work, but tested using the -C command line option.


hlship:
I'm seeing an issue w/ the 0.1.32 version of pretty where it launches a Java application window on the OS X Dock; not sure what it would do on windows or linux. Basically, loading certain AWT classes causes this behavior, and so it needs to be deferred.  Closing until I have a fix.

hlship:
https://github.com/AvisoNovate/pretty/issues/51

hlship:
This is ready for review and merge. I've tested it locally, using the -C option.

hlship:
Ok, I'll work on resolving the fix.  Give me a few minutes.

hlship:
Ok, how's that?

PR #532, opened 2016-11-26
Check :boot.util/omit-stacktrace? to hide stack trace

### LINKED ISSUE #486: Tasks should be able to control how exceptions are shown
In several cases, such as Cljs compilation or test running, for example the stack trace might not be very interesting and is on the way of the interesting message. In these cases the tasks should be able to better decide how to display these exceptions.

Currently tasks need to throw the exception so that Boot detects that the task did not finish and that the next tasks should not be run.

I propose following solution:

Tasks send a special exception using ex-info, e.g. with `:type :boot-exception` or something, with some option keys on what to do, e.g. `:boot-display-exception? false` and optionally the original exception as cause. The default would be to just print the cause.

This could be implemented in `boot.util/print-ex` which is called at least by `watch` task: https://github.com/boot-clj/boot/blob/master/boot/pod/src/boot/util.clj#L253.

Or maybe this logic should be separated to another function, in case the options will in future control other things besides printting the exception. E.g. if the next tasks should run.

Deraen:
Fixes #486 

TODO:

- What should be namespace (and name) of the property? Currently `:boot.util/omit-stacktrace?` but perhaps it should be `:boot.core/omit-stacktrace?` instead?

### DIFF
--- a/boot/pod/src/boot/util.clj
+++ b/boot/pod/src/boot/util.clj
@@ -251,13 +251,17 @@
      (binding [*err* s#] ~@body (str s#))))
 
 (defn print-ex
-  "Print exception to *err* as appropriate for the current *verbosity* level."
+  "Print exception to *err* as appropriate for the current *verbosity* level.
+
+  If ex-data contains truthy :boot.util/omit-stacktrace? value, only exception
+  message is shown."
   [ex]
-  (case @*verbosity*
-    0 nil
-    1 (pretty/write-exception *err* ex nil)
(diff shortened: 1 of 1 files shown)


martinklepsch:
:+1: on `:boot.util/omit-stacktrace?`

alandipert:
Tried it loved it. Thanks :+1:

Deraen:
:+1: @alandipert I loved it too.

PR #537, opened 2016-12-03
Add --mode option to built in target task

ekroon:
For situations where the default 'rw-------' file modes are not enough,
a (posix) mode string can be given to target. All files will get the 
mode written.

When an invalid mode is given, a warning is printed and standard 
permissions will be used.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Changes
 
+## Master
+
+##### Task Options
+- Added `-m, --mode` option to the `target` task &mdash; specifies the file
+  mode for written files &mdash; should only be used when default `rw-------`
+  is not enough.
+
 ## 2.7.0
 
 #### Improved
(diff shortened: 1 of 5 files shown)


alandipert:
Has anyone tried this on Windows?

ekroon:
Good point, I implemented it this way, because tmp files already are created with `PosixFilePermissions`. Maybe I missed a Windows check somewhere as there is a good chance this doesn't work on windows indeed. I did not check, so I will check it and adapt if necessary.

alandipert:
@ekroon cool, thank you.  I haven't checked myself, so it may very well work.  I just wanted to make sure someone has tried it.

alandipert:
Can you please add arities to `link!` and `copy!` that don't take `mode`, but call the arity that does with a value of `nil` for `mode`?

Then the arity change is backward compatible.  Thanks :metal:

alandipert:
It looks like the earlier `PosixFilePermissions` did break Boot on windows, so presumably this would too: https://github.com/boot-clj/boot/issues/541

alandipert:
@ekroon I don't have experience with appveyor but am certainly open to it.

ekroon:
@alandipert took the liberty to also fix #541 on this branch as it is closely related.

PR #539, opened 2016-12-09
Improvements to work with Java 9

tobias:
Changes include:

* update to dynapath 0.2.5 (0.2.4 was broken when AOT'd under < Java 9,
  but used under Java 9)
* conditionally seal the AppClassLoader if it is available (it isn't
  under Java 9)
* seal the new ParentClassLoader from boot-bin
* use a URLClassLoader subclass that exposes .addURL so it can be
  modified (we can't call .setAccessible on URLClassLoader.addURL by
  default under Java 9) as the highest loader in the shim

This is part of the fix; https://github.com/boot-clj/boot-bin/pull/4 is the other half. These changes *shouldn't* break users using a `boot.sh` without PR4. 

Let me know if this (or the approach in PR4) aren't appropriate, and I'm happy to discuss more here or on slack.

### DIFF
--- a/boot/base/src/main/java/boot/AddableClassLoader.java
+++ b/boot/base/src/main/java/boot/AddableClassLoader.java
@@ -1,0 +1,14 @@
+package boot;
+
+import java.net.URL;
+import java.net.URLClassLoader;
+
+// Allows us to have a modifiable ClassLoader without having to call
+// .setAccessible on URLClassLoader.addURL(), since that's not allowed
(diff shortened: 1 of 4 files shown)


tobias:
I should mention that with these changes + PR4, boot works for me under Java 9-ea+148, but I haven't tried anything fancy with it, just starting a repl.

alandipert:
I tested that this doesn't break existing `boot.sh`.  :+1:

PR #540, opened 2016-12-09
Use the new CDN-backed clojars repo by default

tobias:
This also changes the mvn portion of the build to use it as well. Note
that this new repo does not support Java 6. With this change, Java 6
users will have to override the repo manually back to the old value.


martinklepsch:
Thanks @tobias! 🎉 

This should also be added to `CHANGES.md` before a new release is cut.

alandipert:
Boot doesn't run on Java 6 because of NIO, so this is definitely fine. I'll add to CHANGES shortly.

PR #549, opened 2016-12-23
New task: socket-server

### LINKED ISSUE #453: Enabling socket repl via system property causes failure
With the following env vars any boot invocation will fail due to multiple clojure runtimes trying to launch a socket repl on the same port.

```
BOOT_JVM_OPTIONS "-Dclojure.server.repl=\"{:port 5555 :accept clojure.core.server/repl}\""
BOOT_CLOJURE_VERSION "1.8.0"
```

Not sure if this can fixed with code, if not an addition to the docs might be a good idea.

pesterhazy:
Exposes the Socket Server functionality available in Clojure 1.8 or above as a new task. By default, the handler function is clojure.core.server/repl, but a different handler can be specified.

This is useful because Clojure's standard way of enabling the socket server, by setting a system property, doesn't work well with boot (https://github.com/boot-clj/boot/issues/453).

Wheres `boot repl --server`, provides nREPL functionality used in CIDER, `boot socket-server` can be used e.g. with Emacs's inf-clojure mode. It starts up much faster than `boot repl --server` (12s vs 80s), and is more easily accessible using `nc` or `telnet`. But compared to REPLy it is pretty bare-bones and doesn't provide advanced features like auto-completion or beautiful stack traces.

This commit adds a new task rather than extending `repl`. The reason is that Socket Servers provides not just REPL but server functionality in general, and that `repl` semantics are already confusing because of how `--server` and `--client` flags work.

Like `repl`, `socket-server` defaults to picking a port at random and stores it in a file, .socket-port, in the current directory.

Fixes #453

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -479,6 +479,29 @@
           (core/with-post-wrap [_]
             (when (or client (not server)) @repl-cli)))))
 
+(core/deftask socket-server
+  "Start a socket server.
+
+  The default behavior is to serve a simple REPL handled by
+  clojure.core.server/repl. To serve a different handler function, specify a
+  symbol using `--accept'.
+
+  If no bind address is specified, the socket server will listen on 127.0.0.1.
+
+  If no port is specified, an open port will be chosen automatically. The port
+  number is written to .socket-port in the current directory.
+
+  The REPL can be accessed with the command
+
+     $ nc localhost $(cat .server-port)"
+
+  [b bind ADDR      str    "The address server listens on."
+   p port PORT      int    "The port to listen to."
+   a accept ACCEPT  sym    "Namespaced symbol of the accept function to invoke."]
+  (let [repl-soc (delay (repl/launch-socket-server *opts*))]
+    (core/with-pass-thru [fs]
+      @repl-soc)))
+
 (core/deftask pom
   "Create project pom.xml file.
 
--- a/boot/pod/src/boot/repl.clj
+++ b/boot/pod/src/boot/repl.clj
@@ -58,3 +58,21 @@
   [{:keys [bind port init-ns middleware handler] :as options}]
   (let [opts (->> options setup-nrepl-env!)]
     (@start-server opts)))
+
+(defn launch-socket-server
+  "See #boot.task.built-in/socket-server for explanation of options."
+  [{:keys [bind port accept]}]
+  (let [opts {:host (or bind "127.0.0.1")
+              :port (or port 0)
+              :name (gensym "socket-server")
+              :accept (or accept 'clojure.core.server/repl)}]
+    (try
+      (require 'clojure.core.server)
+      (catch java.io.FileNotFoundException e
+        (throw (ex-info "Socket server requires Clojure version 1.8.0 or above"
+                        {:version (clojure-version)}
+                        e))))
+    (let [effective-port (-> ((resolve 'clojure.core.server/start-server) opts)
+                             (.getLocalPort))]
+      (doto (io/file ".socket-port") .deleteOnExit (spit effective-port))
+      (util/info "Socket server started on port %s on host %s.\n" effective-port (:host opts)))))


boot/core/src/boot/task/built_in.clj
```diff
@@ -479,6 +479,28 @@
...
+  The REPL can be accessed with the command
+
+     $ nc localhost $(cat .server-port)"
+
+  [b bind ADDR      str    "The address server listens on."
+   p port PORT      int    "The port to listen to."
+   a accept ACCEPT  sym    "Namespaced symbol of the accept function to invoke."]
+  (let [repl-soc (delay (core/launch-socket-server *opts*))]
+    (core/with-pass-thru [fs]
+      @repl-soc)))
```
martinklepsch: This starts the server but doesn't shut it down afterwards. Not sure if that's necessary, maybe you can comment on that. See `boot.core/cleanup` and `clojure.core.server/stop-server` if you decide to add it.

https://github.com/boot-clj/boot/blob/master/doc/boot.core.md#cleanup

boot/core/src/boot/core.clj
```diff
@@ -1313,3 +1313,19 @@
...
+    (try
+      (require 'clojure.core.server)
+      (catch java.io.FileNotFoundException e
+        (throw (ex-info "Socket server requires Clojure version 1.8.0 or above"
+                        {:version (clojure-version)}
+                        e))))
+    (let [effective-port (-> ((resolve 'clojure.core.server/start-server) opts)
+                             (.getLocalPort))]
+      (doto (io/file ".socket-port") .deleteOnExit (spit effective-port))
+      (util/info "Socket server started on port %s on host %s.\n" effective-port (:host opts)))))
```
martinklepsch: I think this function should be moved into a namespace that is available within pods. 

And could there be a more useful return value than nil? Maybe the name (which is needed when stopping servers) or a function that stops the server?

boot/core/src/boot/task/built_in.clj
Deraen: Cleanup task is currently problematic with REPL. If `boot` is run inside repl task, cleanup from that boot-in-boot would kill also host REPL. So, for now, this probably shouldn't use `cleanup`.

boot/core/src/boot/core.clj
pesterhazy: Updated the PR to return the name

boot/core/src/boot/task/built_in.clj
pesterhazy: I don't know much about boot-in-boot so I'll defer to your judgment that cleaning up is dangerous right now.

boot/core/src/boot/core.clj
pesterhazy: Not sure which namespace would appropriate. I also don't understand the "accessible from pods" distinction - could you elaborate why that would be useful?

martinklepsch: Basically `boot.core` cannot be loaded in a pod because it has stateful stuff that we don't want to replicate. There may be a moment where you want to launch a socket server in a pod though which, with the function in `boot.core`, would require copying the implementation somewhere where it can be loaded by the pod.

For this reason I'm suggesting we put the function somewhere where it is also available in pods. That could be `boot.util` or `boot.pod`, probably there are more, but these two come to mind...

pesterhazy:
Looking into this again, I thought about your suggestion, @martinklepsch. boot.pod or boot.util don't seem appropriate places to put this code as they contain fns to do with pods or utilities.

I don't fully understand code in a pod would want to run `launch-socket-server` as opposed to just calling clojure.core.server/start-server. In my view `launch-socket-server` is just the implementation of the deftask.

I chose core.clj so the fn sits next to `launch-nrepl`. Is part of the concern that we shouldn't clutter core.clj? In that case maybe boot.task.built-in would be better?

martinklepsch:
`launch-nrepl` runs an nrepl server in a pod via `boot.repl/launch-nrepl` 🙂 

The concern is not cluttering core but instead making it available as a general utility (which includes usage from within pods) which is not possible when it's put into core.

Seeing that there is `boot.repl` maybe it should just go there?

pesterhazy:
Would this mean that the code _needs_ to be run in a pod, like the nrepl server?

martinklepsch:
@pesterhazy no, you can still run it in your regular REPL just like that or wherever you want of course (theoretically everything is in a pod but that's more of a detail and probably not what you're concerned about).

pesterhazy:
I've update the PR as per the comments above

martinklepsch:
I think this is good to merge but maybe someone else can take a look too? @Deraen @alandipert?

alandipert:
Just took a look. Awesome, thanks!

Looking at this I thought it might be cool if `launch-socket-server` took an optional pod name argument. Then we'd have a way to get simultaneous pod REPLs. If it took the argument, the pod name should perhaps influence the name of `.socket-port`, or maybe not create such a file at all.

danielsz:
Great work, guys. Thank you!

pesterhazy:
@alandipert, thanks for merging!

Could you outline a scenario where opening a socket server into a pod would be useful?

martinklepsch:
@pesterhazy I *think* what @alandipert is thinking of are situations where you want to "look into a pod" but don't have a reference to the pod object and instead only have the name. 

I have never needed to look into a pod like that myself though.

alandipert:
@pesterhazy getting a REPL in a pod with the `repl` task's `-P` option is useful for developing Boot, and for developing and debugging tasks that use pods.

Using `repl -P` to get a REPL inside a pod is especially useful when tasks don't make their pods public. It's a way to run code in them anyway :sunglasses:

PR #551, opened 2016-12-24
Improve robustness and formatting for `notify` built-in

### LINKED ISSUE #550: `boot notify` on OS X (still) throws IllegalArgumentException
When neither of the options `:uid` or `:title` are passed, the value of `uid` is not set properly from the defaulted `title`, thus still resulting in sometimes passing `nil` as the last argument to `clojure.java.shell/sh`.

I've got a fix for this, that I'll post in a PR soon.  Clearly I never actually ran `boot notify -v` on my Mac before submitting the `notify` task 😱

RadicalZephyr:
There are a number of separate improvements here:

- The fallback behavior for the `uid` key was incorrect (fixes #550)
- Added logic to prevent user passed options from causing a similar error to be thrown from `shell/sh` again
- Added  a median fallback option for MacOS of displaying a notification with AppleScript
- Finally, improved the visual design of the ultimate fallback visual notification system (printing to the console)

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -174,11 +174,11 @@
             (throw t)))))))
 
 (core/deftask ^{:boot/from :jeluard/boot-notify} notify
-  "Aural and visual notifications during build.
+  "Audible and visual notifications during build.
 
-  Default audio themes: system (the default), ordinance, pillsbury,
-  and woodblock. New themes can be included via jar dependency with
-  the sound files as resources:
+  Default audio themes: system (the default), ordinance, pillsbury and
+  woodblock. New themes can be included via jar dependency with the
+  sound files as resources:
 
       boot
       └── notify
@@ -189,8 +189,8 @@
   Sound files specified individually take precedence over theme sounds.
 
   For visual notifications, there is a default implementation that
-  tries to use the `terminal-notifier' program on OS X systems, and
-  the `notify-send' program on Linux systems.
+  tries to use the `terminal-notifier' or `osascript` programs on OS X
+  systems, and the `notify-send' program on Linux systems.
 
   You can also supply custom notification functions via the *-notify-fn
   options. Both are functions that take one argument which is a map of
@@ -200,7 +200,14 @@
   - :type, :file, and :theme.
 
   The visual notification function will receive a map with four keys
-  - :title, :uid, :icon, and :message."
+  - :title, :uid, :icon, and :message.
+
+  The `notify` task always attempts to catch any exceptions that are
+  thrown so as not to cause your build to fail. If no notifications
+  are happening even if you specify `audible` or `visual`, you can see
+  any exceptions that are being caught by changing
+  `boot.util/*verbosity*` (at the command-line use: `boot -v` or `boot
+  -vv`)."
 
   [a audible                    bool      "Play an audible notification"
    v visual                     bool      "Display a visual notification"
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -215,7 +215,8 @@
 
   (let [themefiles (notify/get-themefiles theme (core/tmp-dir!))
         sounds (merge themefiles soundfiles)
-        base-visual-opts {:title (or title "Boot")
+        title (or title "Boot")
+        base-visual-opts {:title title
                           :uid   (or uid title)
                           :icon  (or icon (notify/boot-logo))}
         messages (merge {:success "Success!" :warning "%s warning/s" :failure "%s"}
--- a/boot/core/src/boot/task_helpers/notify.clj
+++ b/boot/core/src/boot/task_helpers/notify.clj
@@ -27,7 +27,7 @@
   (util/warn
    "An exception was thrown while trying to send a notification. To see more info, increase the verbosity of your build with e.g. \"boot -v\" or \"boot -vv\"\n"))
 
-(defn sh-with-timeout [& args]
+(defn- sh-with-timeout [& args]
   (try
     (apply util/dosh-timed 1000 args)
     0
@@ -37,9 +37,15 @@
         (util/print-ex e))
       1)))
 
+(defn- warn-program-not-found [program-name]
+  (util/warn "Could not find the '%s' program on PATH.\n" program-name)
+  false)
+
 (defn- ^{:boot/from :jeluard/boot-notify} program-exists?
-  [s]
-  (= 0 (sh-with-timeout "sh" "-c" (format "command -v %s >/dev/null" s))))
+  [program-name]
+  (or
+   (= 0 (sh-with-timeout "sh" "-c" (format "command -v %s >/dev/null" program-name)))
+   (warn-program-not-found program-name)))
 
 (defn- try-notify-with-shell-program [program-name & args]
   (and
--- a/boot/core/src/boot/task_helpers/notify.clj
+++ b/boot/core/src/boot/task_helpers/notify.clj
@@ -34,6 +34,12 @@
   [s]
   (= 0 (sh-with-timeout "sh" "-c" (format "command -v %s >/dev/null" s))))
 
+(defn- try-notify-with-shell-program [program-name & args]
+  (and
+   (program-exists? program-name)
+   (= 0
+      (apply sh-with-timeout program-name args))))
+
 (defn- escape [s]
   (pr-str (str s)))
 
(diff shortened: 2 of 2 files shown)


boot/core/src/boot/task_helpers/notify.clj
```diff
@@ -26,25 +28,51 @@
   [s]
   (= 0 (:exit (shell/sh "sh" "-c" (format "command -v %s" s)))))
 
+(defn- escape [s]
```
arichiardi: I think `boot.util/pp-str` does exactly this 😀

boot/core/src/boot/task_helpers/notify.clj
```diff
@@ -26,25 +28,51 @@
...
+              (str title)
+              (str message)
+              "--icon"
+              (str icon))
     ((get-method notify-method :default) :default notification)))
 
 (defmethod notify-method :default
   [_ {:keys [message title]}]
-  (printf "%s: %s" title message))
+  (util/info "%s%s %s \u2022 %s\n"
```
arichiardi: :+1:

arichiardi:
Oh also maybe we can squash the changes given that they are all related. Good job!

boot/core/src/boot/task_helpers/notify.clj
arichiardi: Ok maybe it does not escape, I actually never tried this way, sorry for my ignorance, is there a need to explicitely escape a string? Can't we call `pp-str` directly and print to out?

RadicalZephyr: Yeah, so the `escape` function is specifically for use in the new AppleScript fallback.  The reason that I wanted that is because the arguments to the AppleScript function being called should always be quoted strings.  `escape` enforces that no matter what clojure data structure is passed as the message or title, it gets coerced into a string and then has quotes added to the string.  This way the AppleScript being executed will at least always be valid and should do _something_ even if it displays something kind of strange looking.

Does that make sense?  It's the same reason that I started wrapping all values to `shell/sh` in `str`.  No matter what gets passed (even if it's `nil`) the arguments to `shell/sh` will always be a valid way to invoke a program (i.e. only strings) and it won't get confused that something is supposed to be an option (which is the source of the error thrown by #550).

RadicalZephyr:
@arichiardi I squashed the last two commits since they are really the same change, but the other ones I think make sense as their own commits.

Plus, I'm trying to catch up to you on number of commits in boot! 😉  How am I going to do that if all my PRs are one commit?!

boot/core/src/boot/task_helpers/notify.clj
arichiardi: Oh OK that's the reason 😀 No objection to that then

arichiardi:
@RadicalZephyr OK then you are going to compete with four more now 😂
It looks good to me, but I am going to wait if more people want to pitch in before merging.

martinklepsch:
So I ran into this issue with `2.7.1` (not this branch):
```
java.lang.IllegalArgumentException: No value supplied for key: null
```
And I believe this branch fixes it by relying on a quirk in `clojure.java.shell`:

`clojure.java.shell/parse-args` ([source](https://github.com/clojure/clojure/blob/fe0cfc71e6ec7b546066188c555b01dae0e368e8/src/clj/clojure/java/shell.clj#L45)) splits command and options using `(split-with string? args)` which means that if you pass args as strings it will end up with something like this:

```
;; with message being `nil`
'(("terminal-notifier" "-message") (nil "-title" ...))
;; with message being a string
'(("terminal-notifier" "-message" "my message" "-title" ...) ())
```
I think it is intended to pass options via keywords or something (weirdly it doesn't call `name` on these) so that the splitting works properly. I guess in the end it's all joined with spaces so it still works but it still seems kinda wonky to me.

I'm not sure if `clojure.java.shell/sh` is used here for a reason instead of usual `boot.util/dosh` but I'd suggest to replace it just to be consistent. There is one caveat to take care of when using `terminal-notifier` with `boot.util/dosh`:

All these cause indefinite blocking (will investigate why):
```
;; nil values get removed so these are the same
(boot.util/dosh "terminal-notifier" "-message")
(boot.util/dosh "terminal-notifier" "-message" nil)
;; gets executed as "terminal-notifier -message -title"
(boot.util/dosh "terminal-notifier" "-message" nil "-title")
```

I guess it should be relatively simple to sanitize the options so that this doesn't happen.

RadicalZephyr:
You're right, knowing and working around that particular (strange) behavior of `clojure.java.shell` is exactly what fixes it, though fwiw it's also what causes the issue.  I didn't even know about `boot.util/dosh`, I'll switch to using that or `dosh-timed` if #561 gets merged first.

martinklepsch:
@RadicalZephyr Hey :) Are you still planning to update this PR? If you don't have time right now I can also take your changes and make the discussed modifications.

RadicalZephyr:
Sorry, I've been meaning to get to this! I'll take some time right now to update to `dosh-timed`.

RadicalZephyr:
Okay, I put this comment in slack, but I'm going to repeat it here for discussion's sake:

While updating to the new `dosh-timed` I decided to stick with the policy of “never throw from `notify`” which we established in how errors are handled for the audible notifications.  So now errors in the visual notification process are swallowed silently and if anything goes wrong a text notification will be printed in the terminal.

I’m pretty happy with the general idea, but I guess I’m a little unsure about the “swallow all errors silently”.  It seems like it might make it really hard to debug if there’s something weird with the options passed to `notify` that causes it to fail unexpectedly.

Any thoughts?

RadicalZephyr:
Okay, I've updated this to use `dosh-timed` and improved a few other things along the way. Should be ready for another code review 😄

boot/core/src/boot/task/built_in.clj
```diff
@@ -174,11 +174,11 @@
...
 (core/deftask ^{:boot/from :jeluard/boot-notify} notify
-  "Aural and visual notifications during build.
+  "Audible and visual notifications during build.
 
-  Default audio themes: system (the default), ordinance, pillsbury,
-  and woodblock. New themes can be included via jar dependency with
-  the sound files as resources:
+  Default audio themes: system (the default), ordinance and
+  woodblock. New themes can be included via jar dependency with the
+  sound files as resources:
```
martinklepsch: pillsbury is no longer a thing? :(

boot/core/src/boot/task/built_in.clj
```diff
@@ -200,7 +200,14 @@
...
   The visual notification function will receive a map with four keys
-  - :title, :uid, :icon, and :message."
+  - :title, :uid, :icon, and :message.
+
+  The `notify` task always attempts to catch any exceptions that are
+  thrown so as not to cause your build to fail. If no notifications
+  are happening even if you specify `audible` or `visual`, you can see
+  any exceptions that are being caught by changing
+  `boot.util/*verbosity*` (at the command-line use: `boot -v` or `boot
+  -vv`)."
```
martinklepsch: How about logging a warning at default verbosity that there has been an exception? I like that this will never break a build but making exceptions invisible by default seems like it could lead to trouble.

RadicalZephyr: I think at default verbosity level the exception is still actually printed with `print-ex`.  I'm still not super happy with what this says, but I like logging a warning.

Maybe I should take this section out of the help and just log a warning with just: "an exception was thrown while trying to notify, use `boot -v` to see more detail"?

boot/core/src/boot/task/built_in.clj
RadicalZephyr: Pillsbury is totally still a thing. I took it off the list because it was originally not listed and @alandipert mentioned that it was a boot easter egg.  I'm fine keeping it on the list or not, I guess I already spilled the secret :smile:

arichiardi:
@RadicalZephyr @martinklepsch please let's keep it!

`(task-options! notify {:theme "pillsbury"})`

RadicalZephyr:
@arichiardi done :+1:

RadicalZephyr:
@martinklepsch, @arichiardi how does that warning look?

boot/core/src/boot/task_helpers/notify.clj
```diff
@@ -23,12 +23,18 @@
...
+  (util/warn
+   "An exception was thrown while trying to send a notification. To see more info, increase the verbosity of your build with e.g. \"boot -v\" or \"boot -vv\"\n"))
+
 (defn sh-with-timeout [& args]
   (try
     (apply util/dosh-timed 1000 args)
     0
     (catch Exception e
-      (util/print-ex e)
+      (if (= 1 @util/*verbosity*)
```
arichiardi: This is actually always happening because the starting `*verbosity*` is 1. Depending on what you wanted to achieve you could remove it the if or bump it to 2 ;)

RadicalZephyr: Yeah, so that was my thought.  At `*verbosity*` level 1 `print-ex` will actually print the whole stacktrace, which is why I put this `if` in to only print a short warning message that an exception was thrown.  And then, if the `*verbosity*` is set higher or lower the appropriate thing will happen (i.e. at quiet `*verbosity*` nothing will be printed, but at higher levels the full exception will be shown.)

boot/core/src/boot/task_helpers/notify.clj
```diff
@@ -22,29 +23,72 @@
...
+      1)))
+
 (defn- ^{:boot/from :jeluard/boot-notify} program-exists?
   [s]
-  (= 0 (:exit (shell/sh "sh" "-c" (format "command -v %s" s)))))
+  (= 0 (sh-with-timeout "sh" "-c" (format "command -v %s >/dev/null" s))))
+
+(defn- try-notify-with-shell-program [program-name & args]
+  (and
+   (program-exists? program-name)
```
RadicalZephyr: I'm not sure what you mean by return an error-code, but I can definitely see adding a warning when the program is not found, that makes it a bit more discoverable what needs to be installed, though it is also mentioned in the help text.

boot/core/src/boot/task_helpers/notify.clj
arichiardi: Ah, right :) Sorry then this is good

boot/core/src/boot/task_helpers/notify.clj
arichiardi: I mean, yeah, just handle the case gracefully so that we know when there is a program missing

boot/core/src/boot/task/built_in.clj
arichiardi: I like the pattern of logging the warning with the ex msg on normal (no `-v`) execution and printing the stack trace on `-v`.

boot/core/src/boot/task_helpers/notify.clj
RadicalZephyr: Cool, pushed another commit with a warning message

RadicalZephyr:
ping @martinklepsch, @micha or @alandipert 

I think this is ready for a (hopefully :crossed_fingers:) final review when y'all have a moment.

martinklepsch:
I reviewed this and the added commits since I tried it look good to me. Didn't get to building and using locally yet though. Tomorrow maybe 👍

PR #556, opened 2016-12-31
ensure title is not nil when setting default for uid in notify

esp1:
I noticed that the notify task throws an exception when `:visual` is true and `:title` is not explicitly set on Mac OS X. The reason for this is that when the `base-visual-opts` map is constructed, the `:uid` is defaulted to the value of `title`, but if `title` is nil, then `uid` will also be nil. This PR ensures that `title` is not nil when used as the default value for `uid`.

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -215,7 +215,8 @@
 
   (let [themefiles (notify/get-themefiles theme (core/tmp-dir!))
         sounds (merge themefiles soundfiles)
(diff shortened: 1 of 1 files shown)


esp1:
Until this issue is fixed, a workaround is to put `(task-options! notify {:title "Boot"}` in your `.boot/profile.boot`

arichiardi:
I think there is something for this #551, ping @RadicalZephyr

esp1:
@arichiardi  ah cool, i should have checked - it's literally the same commit as here. i'll close this PR and defer to #551 instead. thanks!

Issue #557, opened 2017-01-03
Make  "Commit: adding ..." visible only on -vv


arichiardi:
Proposal: as the title says, it might be a good idea to further increase the level of the numerous:

```
...
Commit: adding   2559f43f090c62f78aa8b658be0fc0ab.1483423236284 main.out/cljs/repl.js.map...
Commit: adding   208b5e74c1791966d2532fe858fbae8a.1453082671000 main.out/weasel/repl.cljs...
Commit: adding   054b0f2728c6d330992e766fb94aedfd.1483423235947 main.out/cljs/spec.js.map...
...
```

Some tasks like `boot-cljs` and `boot-reload` print some stuff on `-v` that at the moment are difficult to read because of the commit messages.

How does it sounds?

Thank you folks :smile:

PR #558, opened 2017-01-03
 Log macros and verbosity level change for commit messages

arichiardi:
As per #557.

### DIFF
--- a/boot/aether/src/boot/aether.clj
+++ b/boot/aether/src/boot/aether.clj
@@ -68,7 +68,7 @@
 
 (defn transfer-listener
   [info]
-  (util/dbug "Aether: %s\n" (with-out-str (pprint/pprint info)))
+  (util/trace* "Aether: %s\n" (with-out-str (pprint/pprint info)))
   (on-transfer info))
 
 (defn ^{:boot/from :technomancy/leiningen} build-url
--- a/boot/pod/src/boot/util.clj
+++ b/boot/pod/src/boot/util.clj
@@ -57,13 +57,53 @@
       (print ((or color identity) (apply format args)))
       (flush))))
 
+(defmacro print**
+  "Macro version of boot.util/print* but arguments are only evaluated
+  when the message will be printed."
+  [verbosity color fmt args]
+  `(when (>= @*verbosity* ~verbosity)
+     (binding [*out* *err*]
+       (print ((or ~color identity) (format ~fmt ~@args)))
+       (flush))))
+
+(defmacro trace*
+  "Tracing macro, arguments are only evaluated when the message will be
+  printed (i.e., verbosity level >= 3)."
+  [fmt & args]
+  `(print** 3 ansi/cyan ~fmt ~args))
+
 (defmacro dbug*
   "Macro version of boot.util/dbug, arguments are only evaluated when the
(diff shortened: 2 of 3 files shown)


boot/pod/src/boot/util.clj
```diff
@@ -57,6 +57,14 @@
       (print ((or color identity) (apply format args)))
       (flush))))
 
+(defmacro print**
+  "Macro version of boot.util/print* but arguments are only evaluated
+  when the message will be printed."
+  [verbosity fmt & args]
+  `(when (>= @*verbosity* ~verbosity)
+     (binding [*out* *err*]
+       (print (ansi/bold-cyan (format ~fmt ~@args))))))
```
arichiardi: Uhm, maybe the color part can be improved here (haven't noticed that immediately, maybe cause it's late :smile:)

arichiardi:
My rigorous testing :smile: 

2017-01-05-224223_114x199_scrot

```
  (boot.util/trace "Fn %s\n" 1)
  (boot.util/dbug "Fn %s\n" 1)
  (boot.util/info "Fn %s\n" 1)
  (boot.util/warn "Fn %s\n" 1)
  (boot.util/fail "Fn %s\n" 1)
  (boot.util/trace* "Macro %s\n" 1)
  (boot.util/dbug* "Macro %s\n" 1)
  (boot.util/info* "Macro %s\n" 1)
  (boot.util/warn* "Macro %s\n" 1)
  (boot.util/fail* "Macro %s\n" 1)
```

micha:
Looks good, thanks @arichiardi! Can we make the trace level output cyan (maybe not bold)? That would help differentiate it from stdout.

arichiardi:
We sure can 😀

arichiardi:
@micha @alandipert Done ;)

arichiardi:
Another idea, should we make `Aether` messages `trace` now?

boot/aether/src/boot/aether.clj
```diff
@@ -68,7 +68,7 @@
 
 (defn transfer-listener
   [info]
-  (util/dbug "Aether: %s\n" (with-out-str (pprint/pprint info)))
+  (util/trace* "Aether: %s\n" (with-out-str (pprint/pprint info)))
```
rfkm: FYI: 
This should fix the bad network throughput. (ref. #565)
In my environment, it takes about 10 minutes to fetch 20MB jar file with `dbug`, while it takes only 20 seconds with `dbug*`. I tried to make a PR by myself but I will wait for this PR to be merged.
BTW, it seems that this code was introduced by [this conflict resolution commit](https://github.com/boot-clj/boot/commit/de16fb099dac2795a5e314038e5edf2d07b92807#diff-b547159e61e59485c59eccc8b7d5f7a1).

Deraen: Yep, macro version is definitively a good call here, else the pprint is called for the info even if dbug prints are disabled.

arichiardi: @rfkm can you by any chance clone and confirm that this solves?

rfkm: @arichiardi I've just tested with this branch and confirmed the `trace*` version is up to 100 times faster than the `dbug` version 👍

Issue #530, opened 2016-11-24
Linux: user limit of inotify watches reached


pesterhazy:
When building my project using CircleCI (Linux), builds fail about 20% of the time with the message

```
+ boot deps
Downloading https://github.com/boot-clj/boot/releases/download/2.5.5/boot.jar...
clojure.lang.ExceptionInfo: User limit of inotify watches reached
    data: {:file "/tmp/boot.user4446813814261296413.clj", :line 5}
       java.io.IOException: User limit of inotify watches reached
sun.nio.fs.LinuxWatchService$Poller.implRegister  LinuxWatchService.java: 264
       sun.nio.fs.AbstractPoller.processRequests     AbstractPoller.java: 260
         sun.nio.fs.LinuxWatchService$Poller.run  LinuxWatchService.java: 329
```

The reason is that boot creates fs watches for the source directories, and CircleCI limits this resource. However, for CI builds watching for changes is not necessary.

One way around this would be to add an option to boot that prevents it from creating the inotify watches, or some way to reduce the number of concurrent watches.

gtrak:
I was pointed here by @alandipert on slack, might be a start :-) https://readmes.numm.org/init/upstart/init/tests/wrap_inotify.c

alandipert:
I made this, but I haven't tested it on CircleCI: https://github.com/alandipert/stubwatch

alandipert:
Mitigated by #564

danielcompton:
I've added docs for this to https://github.com/boot-clj/boot/wiki/Running-Boot-on-CI-systems#circleci.

Deraen:
Also, with CircleCI v2 using Docker containers, there is no such tight limits on inotify watches: https://github.com/cljsjs/packages/blob/master/circle.yml#L6

Deraen:
@danielcompton I added docs for using CircleCI v2: https://github.com/boot-clj/boot/wiki/Running-Boot-on-CI-systems#circleci-v2

vikeri:
@Deraen I still got the Inotify watches error on 2.0 with the boot docker image. So I will add that to the wiki for 2.0 as well

vikeri:
Actually I'm getting the issue sometimes irregardless of if I set `BOOT_WATCHERS_DISABLE: yes`

danielcompton:
@vikeri this isn't in a stable release of boot yet, only in 2.7.2-SNAPSHOT, that might be the problem?

vikeri:
[...]

PR #564, opened 2017-01-06
Allow disabling of inotify.

gtrak:
Workaround for https://github.com/boot-clj/boot/issues/530 - CircleCI inotify exhaustion

Adds a `BOOT_WATCHERS_DISABLE` env variable and property.  

Adds a `--disable-watchers` cli flag that is transferred to a worker pod.

Kills the directory recursion for adding watches based on the value of `boot.util/*watchers?*`.

Tested with the inotify debug script here: https://readmes.numm.org/init/upstart/init/tests/wrap_inotify.c

Repro instructions:
1) compile C file above
```
 $ gcc -c -fPIC wrap_inotify.c -o wrap_inotify.o
 $ gcc wrap_inotify.o -shared -o libwrapinotify.so
```

2) Run boot build with flag and env variable.  INOTIFY_DEBUG will show watches being added.

```
LD_PRELOAD=./libwrapinotify.so INOTIFY_DEBUG=true boot build
``` 

3) Run with flags and env vars and properties.  Note that inotify is enabled but no watches are added in debug output.
```
LD_PRELOAD=./libwrapinotify.so INOTIFY_DEBUG=true boot --disable-watchers build
``` 
```
LD_PRELOAD=./libwrapinotify.so INOTIFY_DEBUG=true BOOT_WATCHERS_DISABLE boot build
```

### DIFF
--- a/boot/core/src/boot/main.clj
+++ b/boot/core/src/boot/main.clj
@@ -32,6 +32,7 @@
                  (update-in %1 [%2] (fnil assoc {}) (keyword k) v))]
    ["-i" "--init EXPR"           "Evaluate EXPR in the boot.user context."
     :assoc-fn #(update-in %1 [%2] (fnil conj []) (read-string %3))]
+   [nil  "--disable-watchers"    "Disable registering file watches (inotify/FSEvents) for constrained environments."]
    ["-f" "--file PATH"           "Evaluate PATH (implies -BP). Args and options passed to -main."]
    ["-h" "--help"                "Print basic usage and help info."]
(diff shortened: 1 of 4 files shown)


gtrak:
When this is working, according to INOTIFY_DEBUG it still initializes the inotify system, but never adds any watches.  I think that should probably be acceptable in terms of file handles, but I'm not sure.  A more invasive/complete approach might be to stub out a mock `IRegister` impl or do the flag check here: `doreg   #(register-recursive %1 %2 [:create :modify :delete])`

danielcompton:
This needs to be added to the changelog I think?

PR #567, opened 2017-01-12
Make pod tests pass, and run them from `make`

bhagany:
Problems with these tests were discovered while working on #566, so I figured I'd take a shot at fixing them.  I elected to change the code as little as possible, and change the tests instead, but I'm not super confident I made the right decision in every case.  Criticism welcome.

### DIFF
--- a/boot/pod/src/boot/file.clj
+++ b/boot/pod/src/boot/file.clj
@@ -21,7 +21,7 @@
 (def ^:dynamic *sync-delete* true)
 (def ^:dynamic *hard-link*   true)
 
-(def windows? (boot.App/isWindows))
+(def ^:dynamic windows? (boot.App/isWindows))
 
 (def tmpfile-permissions
   (into-array FileAttribute
--- a/boot/pod/src/boot/file.clj
+++ b/boot/pod/src/boot/file.clj
@@ -21,7 +21,7 @@
 (def ^:dynamic *sync-delete* true)
 (def ^:dynamic *hard-link*   true)
 
-(def ^:dynamic windows? (boot.App/isWindows))
+(def windows? (boot.App/isWindows))
 
 (def tmpfile-permissions
   (into-array FileAttribute
--- a/boot/pod/test/boot/file_test.clj
+++ b/boot/pod/test/boot/file_test.clj
@@ -1,6 +1,7 @@
 (ns boot.file-test
   (:refer-clojure :exclude [sync file-seq])
   (:require
+    [clojure.string :as str]
     [clojure.test :refer :all]
     [boot.file :as file :refer :all :exclude [name]]
     [clojure.java.io :as io]))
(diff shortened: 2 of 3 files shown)


bhagany:
Hrm, I take it Appveyor runs on Windows...

arichiardi:
Great thanks!

bhagany:
alright, my AppVeyor fix worked :)

arichiardi:
Looks very good to me :)

boot/pod/test/boot/file_test.clj
```diff
@@ -22,7 +22,7 @@
 
 (deftest relative-to-test
   (testing "Nil base"
-    (is (= "out/goog/base.js"          (str (relative-to nil      (io/file "out/goog/base.js"))))))
+    (is (thrown? AssertionError        (str (relative-to nil      (io/file "out/goog/base.js"))))))
```
bhagany: Yeah... `nil` is covered by a `:pre` condition in `relative-to`.  I thought about removing the test, but wasn't sure if there was a good reason for it to be there.  I think it's the only other option though, unless the `:pre` shouldn't be there.

boot/pod/src/boot/file.clj
```diff
@@ -21,7 +21,7 @@
 (def ^:dynamic *sync-delete* true)
 (def ^:dynamic *hard-link*   true)
 
-(def windows? (boot.App/isWindows))
+(def ^:dynamic windows? (boot.App/isWindows))
```
bhagany: This is the change I was most unsure about.  The only reason for `windows?` to be dynamic is for testing purposes, and given that's the case, I didn't think that advertising its dynamic-ness with asterisks was a great idea.

At the time, I didn't know that CI ran on Windows as well, and that fact makes me think there's a better way to do this test? https://github.com/boot-clj/boot/pull/567/files#diff-b6e908691c9bd8d211924e12cfbbcd61R46

arichiardi: I mean it is just a minor thing anyways, given that we don't know all the implications here we might need to wait better opinion.

bhagany: Yeah, I agree.  I'm working on a better opinion myself :)

bhagany:
Okay @arichiardi, I hope you'll like this.  It occurred to me that `io/file` will construct paths with platform-specific separators, so we can rely on AppVeyor tests to test Windows paths without doing it explicitly.  CI passed, so all is well!

I also removed the test for AssertionError, since the `:pre` check for `nil` on `relative-to` makes it redundant.  Let me know what you think.

arichiardi:
Perfect! I think now it looks way way simpler. Maybe the original author of this ns can also chime in. In any case, good job!

bhagany:
@arichiardi I just realized github is still saying that you're requesting changes.  Is that still true?  If so, I'm not sure what they are.

arichiardi:
Uhm no, i am good, let me see if I can cancel the review.

bhagany:
Fantastic, thanks :)

alandipert:
:metal: :guitar: :metal:

Issue #573, opened 2017-01-28
aot should throw exception if no param


mobileink:
The aot task requires one of --all or --namespace.  If neither is provided, either an exception should be thrown or a prominent warning message should be emitted.

esanmiguelc:
This one seems simple enough, I can take a stab at it.

I'm thinking that it should be a warning; something along the lines of "aot will not be run - here's what you should do". Would it make sense for the task to actually just default to `--all`?

mobileink:
I'm inclined to think there should be no default. If the use does not specify what is to be aot-compiled, the task should fail. to avoid surprises.

alandipert:
Closing in favor of #576 where discussion can continue.

PR #576, opened 2017-02-03
Adds check for flags on aot compilation throw when none exist

esanmiguelc:
Addresses #573


alandipert:
Thanks! But I think throwing an exception is a little heavy-handed. How about displaying a warning message instead?

Seems like that clues in new users, while not breaking the build of existing users who might be using the task without knowing it's not doing anything.

Issue #578, opened 2017-02-05
Split an argument in three


arichiardi:
I was trying to have this task option for a task:

```
a parent SYM:VER=PATH       [sym str str]
```

And I got:

```
                              java.lang.Thread.run              Thread.java:  745
java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  617
 java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1142
               java.util.concurrent.FutureTask.run          FutureTask.java:  266
                                               ...                               
               clojure.core/binding-conveyor-fn/fn                 core.clj: 2020
                                 boot.core/boot/fn                 core.clj: 1030
                                               ...                               
                         boot.core/construct-tasks                 core.clj:  986
                                               ...                               
            boot.from.clojure.tools.cli/parse-opts                  cli.clj:  500
   boot.from.clojure.tools.cli/parse-option-tokens                  cli.clj:  350
                               clojure.core/reduce                 core.clj: 6703
                                               ...                               
boot.from.clojure.tools.cli/parse-option-tokens/fn                  cli.clj:  347
                              boot.cli/assoc-fn/fn                  cli.clj:   79
                               boot.cli/parse-type                  cli.clj:   62
                               clojure.core/symbol                 core.clj:  579
java.lang.ClassCastException: clojure.lang.PersistentVector cannot be cast to java.lang.String
  clojure.lang.ExceptionInfo: clojure.lang.PersistentVector cannot be cast to java.lang.String
```

What am I doing wrong?

martinklepsch:
Could you maybe provide a runnable repro snippet @arichiardi?

arichiardi:
@martinklepsch yes definitely, I can actually probably write a small test but I need more time.

This is what fails (in `build.boot`):

```clojure
(deftask split-test
  [a parent SYM:VER=PATH [sym str str]]
  (with-pre-wrap [fs] fs))

```

```shell
$ boot split-test -a abc:efg=2
                              java.lang.Thread.run              Thread.java:  748
java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  624
 java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1149
               java.util.concurrent.FutureTask.run          FutureTask.java:  266
                                               ...                               
               clojure.core/binding-conveyor-fn/fn                 core.clj: 1938
[...]

PR #579, opened 2017-02-05
Handle pom <parent> tag

arichiardi:
* adds support for the `<parent>` tag in `boot.pom` for both writing and reading `pom.xml` files.
* also removes some obsolete file and hooks up boot/worker tests in Makefile.

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -541,16 +541,17 @@
   classifier to your pom.xml, which translates to adding :classifier to this
   task."
 
-  [p project SYM           sym         "The project id (eg. foo/bar)."
-   v version VER           str         "The project version."
-   d description DESC      str         "The project description."
-   c classifier STR        str         "The project classifier."
-   P packaging STR         str         "The project packaging type, i.e. war, pom"
-   u url URL               str         "The project homepage url."
-   s scm KEY=VAL           {kw str}    "The project scm map (KEY is one of url, tag, connection, developerConnection)."
-   l license NAME:URL      {str str}   "The map {name url} of project licenses."
-   o developers NAME:EMAIL {str str}   "The map {name email} of project developers."
-   D dependencies SYM:VER  [[sym str]] "The project dependencies vector (overrides boot env dependencies)."]
+  [p project SYM           sym           "The project id (eg. foo/bar)."
+   v version VER           str           "The project version."
+   d description DESC      str           "The project description."
+   c classifier STR        str           "The project classifier."
+   P packaging STR         str           "The project packaging type, i.e. war, pom"
+   u url URL               str           "The project homepage url."
+   s scm KEY=VAL           {kw str}      "The project scm map (KEY is one of url, tag, connection, developerConnection)."
+   l license NAME:URL      {str str}     "The map {name url} of project licenses."
+   o developers NAME:EMAIL {str str}     "The map {name email} of project developers."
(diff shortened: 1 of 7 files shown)


boot/core/src/boot/task/built_in.clj
```diff
@@ -504,7 +505,8 @@
                :dependencies deps
                :developers developers
                :classifier classifier
-               :packaging (or packaging "jar"))]
+               :packaging (or packaging "jar")
+               :parent (-> parent vec first))]
```
arichiardi: @micha I wanted to have `a parent SYM:VER=PATH       [sym str str]` like in #578, can I do it?

boot/core/src/boot/task/built_in.clj
```diff
@@ -493,7 +493,8 @@
    s scm KEY=VAL           {kw str}    "The project scm map (KEY is one of url, tag, connection, developerConnection)."
    l license NAME:URL      {str str}   "The map {name url} of project licenses."
    o developers NAME:EMAIL {str str}   "The map {name email} of project developers."
-   D dependencies SYM:VER  [[sym str]] "The project dependencies vector (overrides boot env dependencies)."]
+   D dependencies SYM:VER  [[sym str]] "The project dependencies vector (overrides boot env dependencies)."
+   a parent SYM:VER        {sym str}   "The project dependency vector of the parent project"]
```
arichiardi: @micha I wanted to have `a parent SYM:VER=PATH       [sym str str]` like in #578, can I do it?

arichiardi:
I don't remember what I wanted to do in the above comment. It looks like I wanted to split the argument in three (regression aside) but really don't remember what the `path` was for. In any case the regression should in #578 should probably be addressed first.

martinklepsch:
Is the basic feature here still wanted? Is that task option type stuff a blocker?

Lets go ahead and merge this if not. I'll write a changelog entry after merging.

arichiardi:
I think this makes things nicer when writing actual Maven poms for plugins. I have needed this and it does not hurt.
I need to actually check the bug though.

boot/core/src/boot/task/built_in.clj
```diff
@@ -541,16 +541,17 @@
...
+   v version VER           str           "The project version."
+   d description DESC      str           "The project description."
+   c classifier STR        str           "The project classifier."
+   P packaging STR         str           "The project packaging type, i.e. war, pom"
+   u url URL               str           "The project homepage url."
+   s scm KEY=VAL           {kw str}      "The project scm map (KEY is one of url, tag, connection, developerConnection)."
+   l license NAME:URL      {str str}     "The map {name url} of project licenses."
+   o developers NAME:EMAIL {str str}     "The map {name email} of project developers."
+   D dependencies SYM:VER  [[sym str]]   "The project dependencies vector (overrides boot env dependencies)."
+   a parent SYM:VER=PATH   [sym str str] "The project dependency vector of the parent project, path included."]
```
arichiardi: Only this line changed

martinklepsch:
Yikes I think I might have somehow wrecked this PR by using Githubs web interface to address the conflicts in Changes.md.

Merged now though
- Change: https://github.com/boot-clj/boot/commit/e7898a99c9ff7c94f35af7b2fa6a2df997b1f3bf
- Tests: https://github.com/boot-clj/boot/commit/94696f66945b640b4d26f45b699342db55786938
- Changelog: https://github.com/boot-clj/boot/commit/49ea7d6207e694f98375f57776846732e3c91908

PR #581, opened 2017-02-06
Sift now outputs meaningful things on debug

arichiardi:


boot/core/src/boot/task/built_in.clj
```diff
@@ -575,7 +575,7 @@
         action  (partial helpers/sift-action v?)
         process (reduce-kv #(comp (action %2 %3) %1) identity *opts*)]
     (core/with-pre-wrap [fs]
-      (util/info "Sifting output files...\n")
+      (util/dbug* "Sifting with opts: %s\n" (util/pp-str (assoc *opts* :invert v?)))
```
RadicalZephyr: I think personally I would rather still have it print the output by default rather than only on debug.  It's a common behavior for most built-in tasks to print something indicating which task was run, and it helps as a base level sanity check that the tasks you think are running are actually happening.

Big 👍  to making the output more meaningful!

arichiardi:
@RadicalZephyr Ok I have just added the `dbug` entry alongside the `info` one, kind of agree that it would be odd to have it go away all of a sudden

RadicalZephyr:
ummm, @arichiardi I think you pushed the wrong branch to this PR.  It looks like it has your changes for the pom parent tag PR...

arichiardi:
Oh, fixing...Thanks!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

alandipert:
Looks good, thanks! When you have a moment, could you please add to CHANGES.md?

PR #582, opened 2017-02-06
New task: bare-repl

pesterhazy:
This PR adds a new bare-repl task to start a "bare" or terminal repl, as per clojure.main/repl.

A bare repl has a few advantages over the nREPL-based `boot repl` setup:

- fewer moving parts (no pods, no network connection)
- much faster startup
- direct access to tty, i.e. this works with `boot repl -B` but not with `boot repl`: ```(.readPassword (System/console) "Password: " (object-array []))```

Like the normal `repl` task when not invoked with the `-s` option, `boot bare-repl` blocks the boot pipeline until the repl is closed with `^D`.

Essentially the bare REPL is closer to what you get when following Clojure's [Getting Started guide](https://clojure.org/reference/repl_and_main). Disadvantages of `boot bare-repl` compared to `boot repl` include:

- no pretty exceptions
- no readline support (but `rlwrap boot repl -B` works well)

### DIFF
--- a/boot/pod/src/boot/repl.clj
+++ b/boot/pod/src/boot/repl.clj
@@ -59,6 +59,21 @@
   (let [opts (->> options setup-nrepl-env!)]
     (@start-server opts)))
 
+(defn launch-bare-repl
+  [{to-eval :eval,
+    :keys [init init-ns]
+    :or {init-ns 'boot.user}}]
+  (require 'clojure.main 'clojure.repl)
+  ((resolve 'clojure.main/repl)
+   :init (fn []
+           (when init
+             (load-file init))
+           (when to-eval
+             (eval to-eval))
+           (in-ns init-ns)
+           (when (= 'boot.user init-ns)
+             (refer 'clojure.repl)))))
+
 (defn launch-socket-server
   "See #boot.task.built-in/socket-server for explanation of options."
   [{:keys [bind port accept]}]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ ensure that message ends in a newline.
 ##### Tasks
 
 - Added the `socket-server` task for starting a [Clojure 1.8.0+ socket server](https://clojure.org/reference/repl_and_main#_launching_a_socket_server). [#549][549]
+- Added the `bare-repl` task for starting a simple interactive REPL session (a la [clojure.main/repl](https://clojure.org/reference/repl_and_main#_launching_a_repl)) without launching a nREPL server. [#582][582]
 
 #### API Functions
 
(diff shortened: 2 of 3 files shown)


boot/pod/src/boot/repl.clj
```diff
@@ -58,3 +58,8 @@
   [{:keys [bind port init-ns middleware handler] :as options}]
   (let [opts (->> options setup-nrepl-env!)]
     (@start-server opts)))
+
+(defn launch-bare-repl
+  [opts]
```
arichiardi: Should we document the options? https://clojure.github.io/clojure/clojure.main-api.html#clojure.main/repl

boot/pod/src/boot/repl.clj
```diff
@@ -58,3 +58,8 @@
...
+  (require 'clojure.main)
+  ((resolve 'clojure.main/repl)))
```
arichiardi: Probably you have forgotten to pass the options here.

alandipert:
This looks juicy but doesn't merge quickly, can you please give it a look when you have a chance? Thanks :1st_place_medal:

Deraen:
I wonder if this should be completely different task as the options probably differ heavily from nrepl options?

alandipert:
That's a great point.

On Mar 3, 2017 12:18 AM, "Juho Teperi" <notifications@github.com> wrote:

I wonder if this should be completely different task as the options
probably differ heavily from nrepl options?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/boot-clj/boot/pull/582#issuecomment-283893626>, or mute
the thread

pesterhazy:
Based on the feedback by @deraen et al, I've rebased on master and extracted a standalone task, bare-repl. It accepts only the `--init`, `--init-ns` and `--eval` options. What do you think @alandipert?

alandipert:
:100: - Can you please add to CHANGES.md for this one too?

pesterhazy:
@alandipert done

arichiardi:
Sorry if this come out it the blue but these has been a lot of discussion around repls in #unrepl lately and I was thinking that this task could be named `socket` only . The default is to use the accept function that launches the repl but then an option can specify a symbol containing any accept function. We would mimic Clojure's behavior and make thing way more extensible: we have pods, so we could transparently handle the one socket for repl/one or more sockets for tooling.

pesterhazy:
@arichiardi, did you mean to comment on this (already merged) PR: https://github.com/boot-clj/boot/pull/549? This PR is not about the socket server :)

arichiardi:
Lol probably not :+1:

PR #597, opened 2017-03-09
New task: call

pesterhazy:
This task can be used for ad-hoc code, either from the command line or
from build.boot. Use cases:

- run the project's entry point, like `lein run`
- experiments from the CLI: `boot call --print -f clojure-version`
- start a web server: `boot watch cljs call --once -f my.project/start`

As used from build.boot, the call task has the advantage that new users
don't need to learn about with-pass-thru to execute project code.

The `--eval` option allows to execute an arbitrary Clojure from. It is
similar to boot's `--init` but can be run as part of a boot pipeline.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ ensure that message ends in a newline.
 ##### Tasks
 
 - Added the `socket-server` task for starting a [Clojure 1.8.0+ socket server](https://clojure.org/reference/repl_and_main#_launching_a_socket_server). [#549][549]
+- Added the `call` task to execute arbitrary code as part of the pipeline, either via an existing function symbol or by providing a form. Similar to [lein run](https://github.com/technomancy/leiningen#basic-usage), `call` can be used, for example, to start a [component system](https://github.com/stuartsierra/component).
 
 #### API Functions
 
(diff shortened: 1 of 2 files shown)


pesterhazy:
@arichiardi, I've amended the docstring for `--once`

Yes, it is intentional that you can call multiple functions or pieces of code and even mix both, e.g.

```
boot call --print --function clojure-version --function all-ns --eval "(+ 3 4)" --eval "(- 1 5)"
```

Evals are executed after function calls.

RadicalZephyr:
I really like this task.  One thing I would advocate for as well would be an option to call it in the pre section of the task _or_ in the post section.

I recently created a simple task allowing me to do that and used it in a project of mine to stop my custom application system before running the build and `refresh`ing the code.  Then as the post portion of the same task (i.e. after all those other tasks have successfully run), I restart my system with another function call.

So, either another task `call-after` that is nearly identical, or (my preference) options for this task to call the same types of things in the post portion of the task.

kommen:
This task would solve an issue I had recently with starting nodejs for a ClojureScript build target with `watch`, where I then had write my own task using `delay` with `with-pass-truh` to only start node once after after the initial compile.

pesterhazy:
@kommen, your suggestion sounds good, I actually debated whether to add this to the initial pr. I'll add a --post flag when I'm back from vacation.

pesterhazy:
Sorry the last comment was directed at @radicalzephyr

pesterhazy:
As suggested by @RadicalZephyr, I've updated the PR to include a `--post` flag.

alandipert:
I like it. Can you please add to `CHANGES.md`? Thanks!

pesterhazy:
@alandipert I've added the changelog entry and squashed the commits

PR #598, opened 2017-03-09
feat: implement visitFileFailed for mkvisitor

thenonameguy:
Reason:
[This project](https://github.com/degree9/boot-semver/blob/master/src/degree9/boot_semver.clj#L152) does a treewalk on our project structure, which contains folders/files that cannot be opened by the boot process, this causes the process to exit.
This PR fixes this issue by ignoring these files/directories.

### DIFF
--- a/boot/pod/src/boot/tmpdir.clj
+++ b/boot/pod/src/boot/tmpdir.clj
@@ -100,7 +100,11 @@
                    (add-blob! blob path i link)
                    (swap! tree assoc p (map->TmpFile (assoc m :path p :id i :hash h :time t))))
                  (catch java.nio.file.NoSuchFileException _
-                   (util/dbug* "Tmpdir: file not found: %s\n" (.toString p))))) )))))
+                   (util/dbug* "Tmpdir: file not found: %s\n" (.toString p)))))))
+      (visitFileFailed [^Path path ^java.io.IOException e]
+        (with-let [_ fs/skip-subtree]
+          (let [p (str (.relativize root path))]
+            (util/dbug* "Tmpdir: failed to visit: %s\n" (.toString p))))))))
 
 (defn- dir->tree!
   [^File dir ^File blob]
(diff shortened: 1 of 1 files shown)


arichiardi:
Looks good

boot/pod/src/boot/tmpdir.clj
```diff
@@ -100,7 +100,11 @@
                    (add-blob! blob path i link)
                    (swap! tree assoc p (map->TmpFile (assoc m :path p :id i :hash h :time t))))
                  (catch java.nio.file.NoSuchFileException _
-                   (util/dbug* "Tmpdir: file not found: %s\n" (.toString p))))) )))))
+                   (util/dbug* "Tmpdir: file not found: %s\n" (.toString p)))))))
+      (visitFileFailed [^Path path ^java.io.IOException e]
+        (with-let [_ fs/skip-subtree]
+          (let [p (str (.relativize root path))]
```
Deraen: A minor thing: Now that `.relativize` call is outside of the `dbug*` macro call, it is evaluated always. If it was inside the macro call, if would be evaluated only when dbug output is enabled.

Also, it is unnecessary to have both `str` and `.toString` calls.

boot/pod/src/boot/tmpdir.clj
```diff
@@ -100,7 +100,11 @@
...
+            (util/dbug* "Tmpdir: failed to visit: %s\n" (.toString p))))))))
```
Deraen: A minor thing: Now that .relativize call is outside of the dbug* macro call, it is evaluated always. If it was inside the macro call, if would be evaluated only when dbug output is enabled.

Also, it is unnecessary to have both str and .toString calls.

alandipert:
This looks good. @thenonameguy when you have a chance, in addition to the changes @Deraen mentioned, could you please also add to `CHANGES.md`?

Thanks!

martinklepsch:
closing in favor of #629

Issue #555, opened 2016-12-28
Graciously handle empty paths to by-path


arichiardi:
Hi folks!

When the `paths` param to [by-path](https://github.com/boot-clj/boot/blob/2.6.0/boot/core/src/boot/core.clj#L1179) is an empty `seq`, this is the :boom: I receive :

```
                              java.lang.Thread.run              Thread.java:  745
java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  617
 java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1142
               java.util.concurrent.FutureTask.run          FutureTask.java:  266
                             clojure.lang.AFn.call                 AFn.java:   18
               clojure.core/binding-conveyor-fn/fn                 core.clj: 2020
                                 boot.core/boot/fn                 core.clj: 1029
                               boot.core/run-tasks                 core.clj: 1019
              pandeiro.boot-http/eval1138/fn/fn/fn            boot_http.clj:  110
                    boot.task.built-in/fn/fn/fn/fn             built_in.clj:  416
                 boot.task.built-in/fn/fn/fn/fn/fn             built_in.clj:  419
              boot.task.built-in/fn/fn/fn/fn/fn/fn             built_in.clj:  419
                    boot.task.built-in/fn/fn/fn/fn             built_in.clj:  235
               adzerk.boot-reload/eval803/fn/fn/fn          boot_reload.clj:  187
            adzerk.boot-reload/eval803/fn/fn/fn/fn          boot_reload.clj:  188
                        adzerk.boot-reload/changed          boot_reload.clj:   29
                             boot.core/not-by-path                 core.clj: 1266
                        clojure.lang.RestFn.invoke              RestFn.java:  442
                                 boot.core/by-path                 core.clj: 1261
                        clojure.lang.RestFn.invoke              RestFn.java:  442
                          boot.core/file-filter/fn                 core.clj: 1221
                                clojure.core/apply                 core.clj:  657
                       clojure.lang.RestFn.applyTo              RestFn.java:  132
                    clojure.lang.AFn.applyToHelper                 AFn.java:  152
                        clojure.lang.RestFn.invoke              RestFn.java:  399
                       clojure.lang.AFn.throwArity                 AFn.java:  429
```

[...]

PR #608, opened 2017-05-01
Fixes issue #555

DonyorM:
This fixes issue #555

This modifies the file-filter function so that if the criteria seq is
empty, it simply returns an empty list. 

An empty list was chosen instead of simply returning nil because an empty list allows seq functions (`conj`, `cons`, etc.) to continue to work.

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -1218,6 +1218,7 @@
   "A file filtering function factory. FIXME: more documenting here."
   [mkpred]
   (fn [criteria files & [negate?]]
+    (assert (seq criteria) "boot.core/file-filter requires a list of criteria but nil was passed in, make sure your `by-*` calls are passing them.")
     (let [tmp?   (partial satisfies? tmpd/ITmpFile)
           ->file #(if (tmp? %) (io/file (tmp-path %)) (io/file %))
           pred (apply juxt (mapv mkpred criteria))]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 - `set-env!` works even if the user has set `*print-level*` or `*print-length*` to non-nil in their `$BOOT_HOME/profile.boot`. [#587][587] [#586][586]
 - `tmpfile` "Commit: adding..." messages now only appear with `-vv` which eases debugging tasks with `-v` [#557][557]
 - Pod tests pass and can be run with `make` [#567][567]
+- `file-filter` (called by all `by-*` functions) throws an error if no criteria are specified.
 
 #### Fixed
 


boot/core/src/boot/core.clj
```diff
@@ -1218,10 +1218,12 @@
   "A file filtering function factory. FIXME: more documenting here."
   [mkpred]
   (fn [criteria files & [negate?]]
-    (let [tmp?   (partial satisfies? tmpd/ITmpFile)
-          ->file #(if (tmp? %) (io/file (tmp-path %)) (io/file %))
-          pred (apply juxt (mapv mkpred criteria))]
-      (filter-files #(pred (->file %)) files negate?))))
+    (if (empty? criteria)
```
arichiardi: this works, but a more idiomatic way would be:

```
(when (seq criteria))
 ....
````

**Comment by DonyorM** [comment]
Huh, I just noticed that #600 fixed this issue as well. hit023 said for me to do ahead and fix it, and I hadn't thought about checking pull requests. 

Anyway, this does have the advantage of passing the tests.

File: boot/core/src/boot/core.clj (hunk 1 above)
Comment (DonyorM) [inline boot/core/src/boot/core.clj]: Would that not result in a `nil` return if the seq was empty? I'm still new to Clojure and Boot, so I may be missing something.

Comment (arichiardi) [inline boot/core/src/boot/core.clj]: Yep it returns `nil`.
I have been thought that you should always check using `(seq coll)` if the collection contains stuff (wonderfully in The Joy of Clojure). This has to do with how lazy sequences where implemented in Clojure and it is very different from how things are done in other Lisps (you can google `nil punning` and `nil punning next rest in Clojure`).

**Comment by DonyorM** [comment]
@arichiardi Updated it to match your suggestions.

**Comment by arichiardi** [comment]
You know I have been thinking about this and I was wondering whether it is a good idea to filter all the items if no criterias are given...Maybe this is why the behavior was to crash..Is `nil` the same as passing an identity function? Does it need to flip behavior when `negate?` is true?

The least surprising thing is maybe to return all the files in case of empty criteria, maybe, thoughts?

**Comment by DonyorM** [comment]
Though I assume you were asking that as a more general, question, I'll still put my thoughts down.

This is basically a divide by zero situation. Does nothing match anything?

More concretely, my suggest would to be return `nil` following from the divide by zero situation. If you do something impossible, return a result that's basically "undefined." (Ok the analogy is a stretch).

**Comment by arichiardi** [comment]
But ok, so if the behavior is undefined, maybe it is better to throw the error then, this is my biggest doubt.

**Comment by DonyorM** [comment]
Soooo, what do we want to do with this?

**Comment by arichiardi** [comment]
Seeking advice?  😀 The undefined behavior above that you describe is good, but I still think `nil` carries some other connotation. Dunno maybe it is just me, I'd like to hear more opinions.

**Comment by DonyorM** [comment]
I was partially seeking those other opinions :)

**Comment by martinklepsch** [comment]
How about 
```
(assert (seq criteria) "by-path requires a list of matching paths, if you want all files use boot.core/ls")
```
Or something like that. Perhaps added to all the `by-*` functions individually. It seems that passing no critera should essentially be undefined behavior, especially in combination with `:negate?` it can become weird. 

What do you think?

**Comment by arichiardi** [comment]
I came to like the idea of an error actually there, so an assertion makes total sense to me. To all functions in order to specify what's missing. Totally.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

**Comment by arichiardi** [comment]
@DonyorM what do you think? Shall we finish this off? 😁

**Comment by DonyorM** [comment]
@arichiardi @martinklepsch Updated the commit. Let me know if I should ask anything else.

File: boot/core/src/boot/core.clj [hunk 2]
```diff
@@ -1218,6 +1218,7 @@
   "A file filtering function factory. FIXME: more documenting here."
   [mkpred]
   (fn [criteria files & [negate?]]
+    (assert (seq criteria) "file-filet (called from by-*) requires a list of matching paths, if you want all files use boot.core/ls")
```
arichiardi: There is a lil' typo there and I would drop the `if you want ...` because if you call it from `by-ext` is a weird message back...my proposal:

```
boot.core/file-filter requires a list of criteria but null was passed in, make sure your `by-*` calls are passing them.
```

DonyorM: Looks good to me. Just pushed a new commit.

arichiardi:
LGTM

boot/core/src/boot/core.clj
arichiardi: Still don't see it, GitHub seems a bit slow today

DonyorM: Whoops, sorry, somehow didn't commit the changes.

arichiardi:
A part for Changelog and better working ( I am using to pass twice there lol) I think this is finally good to go 😀

DonyorM:
@arichiardi Added changes.md. I put this under improved, since it's kind of new behavior

arichiardi:
Thanks a lot @DonyorM and sorry for this long conversation, I maybe swayed you a bit from the final solution but I see this is an improvement. I spent myself quite some time trying to figure out what the failure was and opened the issue in frustration 😁😁 this will avoid the trouble for other folks 👍 Good job!

DonyorM:
@arichiardi No problem. It wasn't simple to figure out the best way to do this. Thanks for your help!

DonyorM:
Anything else I need to do for this?

arichiardi:
No :+1:  I think @alandipert or @micha will merge it eventually.

martinklepsch:
👍

boot/core/src/boot/core.clj
```diff
@@ -1218,6 +1218,7 @@
   "A file filtering function factory. FIXME: more documenting here."
   [mkpred]
   (fn [criteria files & [negate?]]
+    (assert (seq criteria) "boot.core/file-filter requires a list of criteria but null was passed in, make sure your `by-*` calls are passing them.")
```
martinklepsch: actually `null` could be `nil` here but not too important haha :)

DonyorM: haha, I was stuck in java land there. Just pushed an update to change it.

PR #610, opened 2017-05-04
[Fix] Tmpdir Symbolic Links

burn2delete:
This patch allows adding tmpdir folders which contain symbolic links. Symbolic links are visited by the file visitor, visitfile now checks if file is a symbolic link.


martinklepsch:
@flyboarder would you mind expanding on this change in a Changelog entry?

Outlining previous and with-this-PR behavior in the PR would also be great as I'm not fully aware how Boot dealt with symlinks before.

flyboarder:
I'm not sure this is needed anymore, at the time I was seeing corrupt files being copied around because boot was breaking the symlinks.

Needs more investigation.

PR #612, opened 2017-05-27
Bump pomegranate and dynapath to 1.0.0

arichiardi:
Hello folks!

Now that https://github.com/cemerick/pomegranate/pull/83 is in, I took the time to adapt to `0.4.0-alpha1`. Leiningen is testing the change too and it makes sense to keep in sync.

One notable change is:

> Note that item 3 above is technically a breaking change; however I would argue it is a bugfix for a critical security flaw. It's still possible to explicitly opt-in to non-TLS HTTP repositories if you don't care for security by calling register-wagon-factory! yourself.

I wanted to write it down here so that we don't miss it.

I am going to try it locally of course and report if I find any issue with it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,10 +2,36 @@
 
 ## master
 
+#### Breaking
+
+If you happen to receive "Tried to use insecure HTTP repository without TLS",
+it means your project was configured to download dependencies from a repository
+that does not use TLS encryption.
+This is NOT suppored anymore because it exposes you to trivially-executed
+man-in-the-middle attacks.
+In the rare event that you don't care about the security of the machines
+running your project, you can enable support for unprotected repositories by
+explicitely set a custom `wagon-factory`:
+
+    ;; never do this
+    (require 'cemerick.pomegranate.aether)
+    (cemerick.pomegranate.aether/register-wagon-factory!
+     "http "#(org.apache.maven.wagon.providers.http.HttpWagon.))
+
+It's also possible you have a dependency which includes a reference to an
+insecure repository for retrieving its own dependencies. If this happens it is
+strongly recommended to add an `:exclusion` and report a bug with the
+dependency which does this.
+
+Kudos to the folks working on
+[the related `pomegranate` PR](https://github.com/cemerick/pomegranate/pull/83)
+and `technomancy` for the above explanation.
+
 #### Improved
 
 - Boot is officially Maven Central compatible. Make sure the `sources` and `javadoc` artifacts are on the fileset and `:classifier` is correctly set.
 - Environment variables BOOT_AS_ROOT, BOOT_WATCHERS_DISABLE und BOOT_COLOR accept `true` as a truthy value beside `1` and `yes` [#631][631]
+- Bump [pomegranate](https://github.com/cemerick/pomegranate) and [dynapath][https://github.com/tobias/dynapath] to `1.0.0`. [#612][612]
 
 #### Fixed
 
@@ -20,6 +46,7 @@
 [654]: https://github.com/boot-clj/boot/issues/654
 [566]: https://github.com/boot-clj/boot/pull/566
 [631]: https://github.com/boot-clj/boot/issues/631
+[612]: https://github.com/boot-clj/boot/pull/612
 
 ## 2.7.2
 
--- a/boot/aether/project.clj
+++ b/boot/aether/project.clj
@@ -16,5 +16,6 @@
   :dependencies [[org.clojure/clojure               "1.6.0"  :scope "compile"]
                  [boot/base                         ~version :scope "provided"]
                  [boot/pod                          ~version :scope "compile"]
-                 [com.cemerick/pomegranate          "0.3.1"  :scope "compile"]
-                 [org.apache.maven.wagon/wagon-http "2.9"    :scope "compile"]])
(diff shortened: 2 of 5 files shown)


boot/aether/project.clj
```diff
@@ -16,5 +16,5 @@
   :dependencies [[org.clojure/clojure               "1.6.0"  :scope "compile"]
                  [boot/base                         ~version :scope "provided"]
                  [boot/pod                          ~version :scope "compile"]
-                 [com.cemerick/pomegranate          "0.3.1"  :scope "compile"]
-                 [org.apache.maven.wagon/wagon-http "2.9"    :scope "compile"]])
+                 [com.cemerick/pomegranate          "0.4.0-alpha1"  :scope "compile"]
+                 [org.apache.maven.wagon/wagon-http "2.12" :exclusions [org.apache.maven.wagon/wagon-provider-api]]])
```
Deraen: wagon-http is missing `:scope "compile"`

arichiardi: Ok will add it

arichiardi:
Probably worth adding some entry to the wiki as well, see https://github.com/technomancy/leiningen/commit/809c7d33ea5ef2ffefa1dc31ace839155955600d

arichiardi:
Just wanted to report that it is not breaking anything. I have also tried the `:repositories` key with a private Nexus with no issues.

Deraen:
Is this up-to-date with Pomegranate 0.4.0 final?

arichiardi:
Yeah no, and I think I should have a look and see if this needs more bumping

martinklepsch:
@arichiardi can you bump the dependency and add a changelog entry mentioning the need for `register-wagon-factory!` when using non-TLS HTTP repos?

martinklepsch:
Thanks @arichiardi. I've noticed that pomegranate `1.0.0` has been released ([changelog](https://github.com/cemerick/pomegranate/blob/master/CHANGES.md)) with Java 9 support. 

Wondering if anyone has opinions on how we should upgrade to that? The new pomegranate also depends on dynapath 1.0.0 which we probably probably should upgrade to as well.

Both of these releases are marked as BREAKING but if I understand [this section](https://github.com/tobias/dynapath#note-on-urlclassloader) on the dynapath README correctly we might be lucky and @tobias already addressed those breakages in Boot.

seancorfield:
Is there also perhaps an argument in favor of shifting to the tooling that `tools.deps` uses (Apache Resolver?) or am I missing some element of what pomegranate provides that is fundamentally different?

martinklepsch:
@seancorfield I think the big difference between these two is that Pomegranate allows modifying the classpath, something that isn't really a goal with tools.deps as far as I can tell. Might still be worth discussing using it for dependency resolution though.

arichiardi:
I wonder what breaking means in that case... probably I should just try to bump

tobias:
@martinklepsch correct, `boot` can ignore the breaking changes from dynapath, since it doesn't use a `URLClassLoader` since https://github.com/boot-clj/boot/commit/a046a497a8bb7f3d1e7aa8d4db4a81c51beaef7d and https://github.com/boot-clj/boot-bin/commit/172a3d2b12c8299b0bac7f1431f8a621c695e5ff

martinklepsch:
Awesome, so that means we can just upgrade to Pomegranate 1.0.0 / Dynapath 1.0.0. Thanks for already taking care of that @tobias, much appreciated!

martinklepsch:
@arichiardi With that in mind what do you think about the following:
- one commit that upgrades to 0.4.0
- one commit that upgrades to 1.0.0

This way we can git-bisect any differences between 0.4.0 and 1.0.0
Changelog entry could be done in both commits, just updating the version in the second.

If you think the two commits are not worth it just make one 🙂

arichiardi:
Ops I think I have already pushed on this branch. Not too much change from 0.4.0 anyways...but let me update the commit message

arichiardi:
This is working fine on my end, reject things when necessary (private Nexus with no `https` enabled)

CHANGES.md
```diff
@@ -2,10 +2,36 @@
...
+
+Kudos to the folks working on
+[the related `pomegranate` PR](https://github.com/cemerick/pomegranate/pull/83)
+and `technomancy` for the above explanation.
+
 #### Improved
 
 - Boot is officially Maven Central compatible. Make sure the `sources` and `javadoc` artifacts are on the fileset and `:classifier` is correctly set.
 - Environment variables BOOT_AS_ROOT, BOOT_WATCHERS_DISABLE und BOOT_COLOR accept `true` as a truthy value beside `1` and `yes` [#631][631]
+- Bump [pomegranate](https://github.com/cemerick/pomegranate) and [dynapath][https://github.com/tobias/dynapath] to `1.0.0`. [#612][612]
```
martinklepsch: Should be `[dynapath](https://github.com/tobias/dynapath)` 🙂

martinklepsch:
Merged via https://github.com/boot-clj/boot/commit/c5e5651e89e85192dc9dcb4f2bb91572bae6367a & https://github.com/boot-clj/boot/commit/f26a12c7510360709746d08bd24301cf11dbf70d with the Markdown link fix mentioned above. 

We should probably release a SNAPSHOT soon so people can help testing this without building locally. Also we should release SNAPSHOTs more often :)

PR #613, opened 2017-05-28
Fix #547 - Concatenate strings and load the result, rather than reading each separately

### LINKED ISSUE #547: Boot as shell silently quits when the script is malformed
Here's a valid `test.clj`:

```
(println "Loading")

(defn -main [& args]
  (println "Hello" args)))
```

When launched with `boot -f test.clj` it works correctly:

```
Loading
Hello (1 2 3)
```

But if I put an extra paren after the first form:

```
(println "Loading")) ;; <-- HERE

(defn -main [& args]
  (println "Hello" args)))
```

When running the same script I get:

```
Loading
```

And the exit code is zero. The expected behavior would be to signalize that the script is malformed.

bhagany:
I've verified that @alexander-yakushev's example correctly fails because of the unmatched delimiter, as well as verifying that aliased namespaced keywords work correctly.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ escape `%` in message to prevent errors about bad string formatting, and
 ensure that message ends in a newline.
 - Artifact upload slow because of an expensive evaluation of a debugging arguments for all calls to `transfer-listener` [#565][565] [#558][558]
 - With-cp does not consider source/resource paths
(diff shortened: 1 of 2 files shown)


bhagany:
I should probably also mention that the temp file is no longer written, and the code is loaded directly from the string, following [this comment](https://github.com/boot-clj/boot/issues/547#issuecomment-268857058)

PR #618, opened 2017-06-16
Fixes #617: Add simple template processor in `jar :file`.

### LINKED ISSUE #617: [new feature] Jar :file task would treat an input string as template
`boot.task.built-in/jar` uses the argument `:file` value literally. It would be useful if the value was treated as a template. For exmaple "{project}-{version}-standalone.jar".

jgrzebyta:
- accepts `version` and `project` templates 

Signed-off-by: Jacek Grzebyta <jgrzebyta@users.noreply.github.com>

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -829,7 +829,7 @@
 (core/deftask jar
   "Build a jar file for the project."
 
-  [f file PATH        str       "The target jar file name."
+  [f file PATH        str       "The target jar file name template. Patterns '{project}' and '{version}' are replaced by relevant values."
    M manifest KEY=VAL {str str} "The jar manifest map."
    m main MAIN        sym       "The namespace containing the -main function."
    p project SYM      sym       "The project symbol -- used to find the correct pom.xml file."]
@@ -846,7 +846,8 @@
                 (boot.pom/pom-xml-parse-string ~(slurp pom))))
             pomname (when (and project version)
                       (str (name project) "-" version ".jar"))
-            fname   (or file pomname "project.jar")
+            file-parsed-string (string/replace file #"\{(\w+)\}" (fn [[_ group]] ((keyword group) {:project (name project) :version version})))
+            fname   (or file-parsed-string pomname "project.jar")
             out*    (io/file tgt fname)]
         (when (and project (seq p))
           (util/warn "Multiple pom.xml files for project: %s\n" project))


Deraen:
Lein uses `format` for these, and only supports using `version`: `:jar-name "hello-%.jar"`.
Maven supports properties: `<file>foo-${foo.version}</file>`.

I think copying Lein behaviour would be sensible, unless there is a good reason to also support project name?

jgrzebyta:
What about making separate - uberised - jar? In maven user might compile
normal jar or the executable '-with-dependencies' one. Does your solution
require having separate build file with different project name?

Best regards,
Jacek

alandipert:
I think we're not really comparable to either lein or maven in this case, since we're programmable. The user can easily fill a template string in their build.boot based on strings they define and pass as arguments to tasks, which could also be the strings used to generate the pom.

That said, I see no reason not to do this, even if its arguably going against the grain in our model. It's backward compatible and can save some typing. :+1: 

@Deraen if you are OK with it too please feel free to merge when you see this.

boot/core/src/boot/task/built_in.clj
```diff
@@ -846,7 +846,8 @@
                 (boot.pom/pom-xml-parse-string ~(slurp pom))))
             pomname (when (and project version)
                       (str (name project) "-" version ".jar"))
-            fname   (or file pomname "project.jar")
+            file-parsed-string (string/replace file #"\{(\w+)\}" (fn [[_ group]] ((keyword group) {:project (name project) :version version})))
```
martinklepsch:
```
(string/replace nil ,,,)
```
This will throw, no?

martinklepsch: The replacement function seems a bit hard to read and will throw if an unknown pattern is provided. 
Maybe something like the following could be used?

```clojure
(fn [[_ group]]
  (case group 
    "project" (name project)
    "version" version
    (do (util/warn "Unknown pattern {%s} in jar task's :file option\n" group)
        (str "{" group "}")))))
```
This will provide a helpful message if the user makes a typo instead of this:
```
NullPointerException   java.util.regex.Matcher.quoteReplacement (Matcher.java:701)
```

For readability the replacement function could also be put into it's own local var or even in a top-level `defn-`.

martinklepsch:
@jgrzebyta are you planning to get back to this or should it be closed? 🙂

jgrzebyta:
@martinklepsch I will close this PR.

martinklepsch:
Ok! Thanks for your efforts! 💛

PR #621, opened 2017-06-20
Improve push and sign-jar documentation

arichiardi:


martinklepsch:
Not 100% sure if we need it in both places but looks fine besides that minor point 👍

PR #622, opened 2017-06-21
Document classifier

arichiardi:
Debug logs for install and mention :classifier in docstrings


martinklepsch:
I don't really understand what these docstrings mean :D (But I also don't know a whole lot about classifiers.. :))

arichiardi:
@alandipert This should probably go in as well for completeness :smile:

Issue #623, opened 2017-06-21
Boot does not sign jars with classifiers


arichiardi:
# Boot Bug Report

## Platform details

Platform:
*Linux*

Platform version:
```
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-0ubuntu1.16.04.2-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
```

## Boot details

Boot version: master
`build.boot` present? (yes/no): yes
`~/.boot/profile` present? (yes/no): no
Task name? (if applicable): `push`

### `build.boot` content (if applicable)

```clojure
;; build.boot content here
```

### `~/.boot/profile` content (if applicable)

```clojure
... push {:pom (str extractor-project)
               gpg-sign true
               ...
              }
```

## Description

The gpg signature produces wrong output file names for artifacts with the `:classifier` tag in their coordinates, which might be present if the deployed jar is of the `-sources` or `-javadoc` kind.

## Steps to reproduce
Just try to `boot push --gpg-sign` three files `main-0.1.0.jar`,  `main-0.1.0-sources.jar` and `main-0.1.0-javadoc.jar`.
The only generated `.asc` file will be `main-0.1.0.jar.asc`.

The bug seems in https://github.com/boot-clj/boot/blob/ac835cea8c18e926bb843e01579b86e3f30458d2/boot/pod/src/boot/gpg.clj#L52 which does not consider `:classifier`.
Will debug later.

PR #624, opened 2017-06-21
Use classifier when deploying artifacts

arichiardi:
This patch fixes bug #623.

This also refactors the code so that if a classifier is found in the artifact definition, we don't sign the pom. The pom is signed only for the main .jar artifact.

Now with `:classifier` set correctly to either `javadoc` and `sources`, we can deploy signed artifact for Maven Central/Sonatype. Before, the artifacts were rejected by their automated "quality assurance" system.

### DIFF
--- a/boot/pod/src/boot/gpg.clj
+++ b/boot/pod/src/boot/gpg.clj
@@ -50,7 +50,7 @@
     (str file ".asc")))
 
 (defn sign-jar
-  [outdir jarfile pompath opts]
+  [outdir jarfile pompath gpg-options]
   (shell/with-sh-dir
     outdir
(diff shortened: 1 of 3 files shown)


boot/pod/src/boot/gpg.clj
```diff
@@ -59,7 +59,7 @@
                     (.deleteOnExit)
                     (spit (pod/pom-xml jarfile pompath)))
           pomout  (io/file outdir (.replaceAll jarname "\\.jar$" ".pom.asc"))
-          sign-it #(slurp (sign (.getPath %) opts))]
```
arichiardi: forgive me if I did this, but is is clearer

Issue #574, opened 2017-01-30
deftask list syntax


aaronlahey:
I prefer to define my single-arity functions the same way I do multi-arity functions. Example:

```clj
(defn foo
  ([]
   (println "foo")))
```

I did this out of habit on the `deftask` macro...

```clj
(deftask package
  "Example task..."
  ([]
   (comp (jar)
         (target))))

```

...and got this error which was a bit confusing:

```
 clojure.lang.ExceptionInfo: Wrong number of args (1) passed to: cli/argspec->cli-argspec
    data: {:file
           "/var/folders/mf/4t4byj3s7xncnj8lpr_1tzf40000gn/T/boot.user2925592664059683955.clj",
           :line 15}
clojure.lang.ArityException: Wrong number of args (1) passed to: cli/argspec->cli-argspec
               ...
boot.main/-main/fn   main.clj: 196
   boot.main/-main   main.clj: 196
               ...
  boot.App.runBoot   App.java: 399
     boot.App.main   App.java: 488
               ...
         Boot.main  Boot.java: 258
```

Is there any interest in making `deftask` a bit more like `defn` (supporting a list containing the bindings/forms) or adding a more helpful error message?

arichiardi:
My 2c: it is interesting but I guess also kind of hard cause if you open that Pandora's box then you also need to parse multi-arity task options.

And what happens if I use I have one `deftask` with two arities with different sets of options?

I guess that would be the most difficult case to handle correctly and frankly I see no big benefit in it. Of course I might be totally wrong here 😀😀

RadicalZephyr:
I think adding a more helpful error message would be great, but I don't think adding support for multiple arities makes any sense because boot tasks are already fundamentally multi-arity because all options are passed as keyword-style args and most should probably be optional or have sane-defaults.

Also, along with what @arichiardi points out, there are other things (`task-options!`, help text generation that would be vastly complicated by such a change without any real benefit.

Better error messages are always a good thing though :+1:

aaronlahey:
Thanks for the input @arichiardi and @RadicalZephyr!

For the record, I wasn't suggesting adding support for multiple arities, just adding support for...

```clj
(deftask foo
  ([]
   (comp (a) (b))))
```

...in addition to...

```clj
(deftask foo
  []
  (comp (a) (b)))
```

...and perhaps throwing an exception stating "`deftask` does not support multiple arities" for something like...

```clj
(deftask foo
  ([]
   (comp (a) (b)))
[...]

PR #627, opened 2017-06-27
Improves error message when user tries to create multi-arity tasks.

DonyorM:
Fixes issue #574.

When a user tries to use a mult-arity format for deftask, it now throws an `IllegalArgumentException` with the error message: "Multiple arity format not supported for tasks. Use single arity format. Ex (deftask build [x y ...] ;commands)"


DonyorM:
@arichiardi Whoops, fixed that.

DonyorM:
Anything else I need to do on this?

PR #629, opened 2017-07-11
Implement visitFileFailed for mkvisitor

martinklepsch:
addressing Juho's feedback and adding an entry to the changelog /via #598


martinklepsch:
Just rebased against master.

PR #634, opened 2017-07-21
fix #633; also search for pom and jar files in fileset

### LINKED ISSUE #633: push tasks :file and :pom options don't search in fileset
The code to find the jar and pom file given to a `push` tasks uses `io/file` which essentially prevents you from specifying a file in the fileset. This in turn requires you to output things into `target` to make things work.

Proposal: We search for both files in the following order:

- exact fileset match
- classpath match (io/resource... )
- file match (io/resource ...)

Also exceptions around some of this stuff could be improved. If there's a realistic chance to get this into 2.7.2 I'd be very happy to start working on this ASAP.

martinklepsch:
Not sure if this is the best way to handle it. 

- Fileset matches take priority. 
- Classpath is not checked explicitly (in the worker pod I could not find the pom.xml through the classpath).

### DIFF
--- a/boot/core/src/boot/task/built_in.clj
+++ b/boot/core/src/boot/task/built_in.clj
@@ -1003,19 +1003,25 @@
   (let [tgt (core/tmp-dir!)]
     (core/with-pass-thru [fs]
       (core/empty-dir! tgt)
-      (let [jarfiles (or (and file [(io/file file)])
+      (let [jarfiles (or (and file [(if-let [fs-jar (get-in fs [:tree file])]
+                                      (core/tmp-file fs-jar)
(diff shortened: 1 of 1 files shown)


alandipert:
We're kind of running out of argument letters, but adding a new argument to do this seems less breaky to me. Even if it's an obscure letter, it would at least have its own doc string and there'd be no ambiguity between file in fileset or relative path on disk.

What do you think?

martinklepsch:
Closing this issue as the problem should probably be solved [by changing the uberjar exclusions pattern](https://github.com/boot-clj/boot/issues/633#issuecomment-317009904).

PR #652, opened 2017-10-25
Display ex-info data when omitting stacktrace

seancorfield:
If you throw with `:boot.util/omit-stacktrace?`, you only get the message displayed -- you lose any other data in the exception.

This change will show the `ex-info` data (without `:boot.util/omit-stacktrace?`) if anything beyond `:line` was present. `:line` is injected by `boot.main` and can be suppressed if no other data was added by the user.

### DIFF
--- a/boot/pod/src/boot/util.clj
+++ b/boot/pod/src/boot/util.clj
@@ -328,7 +328,13 @@
   [ex]
   (cond
     (= 0 @*verbosity*) nil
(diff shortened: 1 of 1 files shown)


Deraen:
I think this is outside of intended omit-stacktrace use. Tasks that use the option can add all the necessary information to the message, or print addition data themselves.

Issue #654, opened 2017-10-30
Can't add clojure source jar to classpath


pdbrown:
Hi! I just started using boot, and am really liking it -- especially the order it brings to my byzantine frontend builds. The ability to boot a repl from which I can hack around in both boot and the rest of my project is pretty cool too. I noticed this little issue as I was migrating from a lein project, and will send a PR.

# Boot Bug Report

## Platform details

Platform: Linux
Platform version: Debian 4.9.30-2+deb9u2
JRE/JDK version: openjdk 1.8.0_131

## Boot details

Boot version: 2.7.2
`build.boot` present?: yes
`~/.boot/profile` present?: no

### `build.boot` content

```clojure
(set-env! :dependencies
          #(conj % '[org.clojure/clojure "1.9.0-beta2" :classifier "sources"]))
```

### `boot.properties` content

```
BOOT_CLOJURE_NAME=org.clojure/clojure
BOOT_CLOJURE_VERSION=1.9.0-beta2
BOOT_VERSION=2.7.2
```

## Description
One can't currently add clojure java sources to the classpath, and therefore can't jump to clojure java definitions in CIDER. When adding a dependency on, say, `[org.clojure/clojure "1.9.0-beta2" :classifier "sources"]`, the corresponding jar isn't added to the classpath because it's filtered out by `boot.pod/resolve-dependency-jars`. Other source jars, which aren't caught by that filter, can be added without issue.

## Steps to reproduce
1. Open (the above) `build.boot` in CIDER
2. Run `cider-jack-in`
3. Notice the following does not contain the source jar:
```clojure
(->> (clojure.string/split (:boot-class-path (get-env)) #":")
     (filter #(clojure.string/includes? % "org/clojure/clojure")))
```

martinklepsch:
Closed via #655

PR #655, opened 2017-10-30
Allow clojure source jars onto the classpath (#654)

pdbrown:
The rm-clj filter of boot.pod/resolve-dependency-jars prevents adding the
org.clojure/clojure dependency to work around an old issue (see 61c948f),
but was also keeping clojure source artifacts off the classpath.
This change narrows the filter to remove only clojure dependencies that aren't
sources, so one can add e.g.
[org.clojure/clojure "1.9.0-beta2" :classifier "sources"]
to enable jumping to clojure java sources in CIDER.

### DIFF
--- a/boot/pod/src/boot/pod.clj
+++ b/boot/pod/src/boot/pod.clj
@@ -593,7 +593,8 @@
   (let [clj-dep (symbol (boot.App/config "BOOT_CLOJURE_NAME"))
         rm-clj  (if-not ignore-clj?
                   identity
-                  (partial remove #(= clj-dep (first (:dep %)))))]
+                  (partial remove #(and (= clj-dep (first (:dep %)))
+                                        (not-any? #{[:classifier "sources"]} (partition 2 (:dep %))))))]
     (->> env resolve-dependencies rm-clj (map (comp io/file :jar)))))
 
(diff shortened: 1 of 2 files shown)


pdbrown:
Addresses https://github.com/boot-clj/boot/issues/654

boot/pod/src/boot/pod.clj
```diff
@@ -593,7 +593,8 @@
   (let [clj-dep (symbol (boot.App/config "BOOT_CLOJURE_NAME"))
         rm-clj  (if-not ignore-clj?
                   identity
-                  (partial remove #(= clj-dep (first (:dep %)))))]
+                  (partial remove #(and (= clj-dep (first (:dep %)))
+                                        (not-any? #{[:classifier "sources"]} (partition 2 (:dep %))))))]
```
Deraen: Maybe `(not= "sources" (:classifier (into {} (:dep %))))` is simpler/easier to understand?

pdbrown: It looks like `into` wants pairs it can conj onto the map. We could say
```clojure
(not= "sources" (:classifier (into {} (map vec (partition 2 (:dep %))))))
```
I'd stick with the original, but up to you!

Issue #641, opened 2017-09-20
OSX: Reasonably large number of directories causes "FSEventStreamStart: register_with_server: ERROR"


minikomi:
# Boot Bug Report

## Platform details

* Platform (macOS, Linux, Windows): Mac OSX 
* Platform version: Sierra 10.12.6 
* JRE/JDK version: "1.8.0_121"

## Boot details

```
BOOT_CLOJURE_NAME=org.clojure/clojure
BOOT_CLOJURE_VERSION=1.8.0
BOOT_VERSION=2.7.1
```

### `build.boot` content (if applicable)

```clojure
(set-env!
 :resource-paths #{"resources"})
```

## Description

When there are a reasonably large number of directories in `resources`, running `boot`, even without any tasks, causes many errors to the tune of:

```
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-09-20 11:45 java[9766] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
[...]

PR #663, opened 2017-11-23
Use bootignore patterns in watcher workers

yannvanhalewyn:
Boot uses the ignore patterns parsed from `.bootignore` when syncing src and resource directories, but not in the underlying watch processes. This PR offers a start to solving this discrepancy.

It is also a potential solution for people have node-modules / figwheel builds et ceterae in resource directories experiencing FSEvent dropouts on Mac OSX (eg: https://github.com/boot-clj/boot/issues/641) and improves performance in those cases - but mainly feels consistent.

**note** This won't exclude any added files after the watcher started, since the watchservices take dirs and don't have ignore options. It seems like that could be fixed in `boot.file/watcher!`. This also won't extend the `watch` task yet. It only will filter out .bootignore matches at register time when boot boots.

**disclaimer** I didn't test the entire build since `make install` fails with a TLS error. I just did not want to spend time debugging the build tool for debugging the build tool I was debugging. I did however test the adapted functions in a repl.

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -741,7 +741,8 @@
         (let [q       (LinkedBlockingQueue.)
               watcher (apply file/watcher! :time dirs)
               paths   (into-array String dirs)
-              k       (pod/with-invoke-worker (boot.watcher/make-watcher q paths))]
+              k       (pod/with-invoke-worker
+                        (boot.watcher/make-watcher q paths :ignore @bootignore))]
           (daemon
             (loop [ret (util/guard [(.take q)])]
               (when ret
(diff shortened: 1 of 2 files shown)


arichiardi:
Thanks, seems sensible

alandipert:
This looks good, thank you for your contribution! However it needs a CHANGES.md entry.

martinklepsch:
@yannvanhalewyn Hey, would you mind adding an entry to CHANGES.md?

martinklepsch:
> I didn't test the entire build since make install fails with a TLS error. I just did not want to spend time debugging the build tool for debugging the build tool I was debugging. I did however test the adapted functions in a repl.

@yannvanhalewyn could you check if you can still reproduce this and open an issue with the full exception/error? Thanks a lot in advance! I hope we can reduce these kinds of issues to make contributing easier.

PR #666, opened 2017-12-12
add a warning when the repl tasks :eval option is passed a string

martinklepsch:
Currently this works:
```
boot repl -c -e '(println "hello")'
```

but this doesn't despite being the obvious "translation" of the above.
```clojure
(deftask connect
  []
  (repl :client true
        :eval "(println \"hello\")"))
```

What's expected in this case is a quoted form; a string is just evaluated as a string essentially not doing anything:
```clojure
(deftask connect
  []
  (repl :client true
        :eval '(println "hello")))
```

The change in this PR prints a warning if a user passes a string instead of a quoted form in Clojure code / build.boot:
```
When passing :eval to the repl task in your build.boot, use a quoted form instead of a string
```

I considered just using `read-string` but a warning seems fair and might also educate the user about the differences when passing `edn` options via the CLI vs. Clojure code.

Considered adding a Changelog entry but didn't really seem interesting enough.


martinklepsch:
Thanks for the review @arichiardi! 

I think ultimately we should consider a `form` option type that will only accept Clojure forms a la `(init)` and not accept anything else (i.e. as in this case: edn).

PR #669, opened 2017-12-21
document release process in CONTRIBUTING.md

martinklepsch:
- [x] document what file needs to be uploaded to GitHub


martinklepsch:
Oh, didn't know about this! Does it make a difference if we build with Oracle JDK or OpenJDK (assuming both v7)?

martinklepsch:
@alandipert I'll merge this now because I want to update in other PRs, if you want to add add a boot-build-box link + some context, go for it.

alandipert:
I don't think it matters

Alan

PR #670, opened 2017-12-21
updated to pretty 0.1.34

flyboarder:


martinklepsch:
Wondering if this should be mentioned in the changelog? What do you think?

PR #678, opened 2018-01-02
digest jar file instead jar path

aJchemist:
In the original source, files extracted from the old jar are cached, preventing you the new jar file contents from adding to fileset, Even if the artifact is changed.


martinklepsch:
@aJchemist hey, sorry for just noticing now, would you mind adding a line to the changelog file?

aJchemist:
@martinklepsch ping!

PR #679, opened 2018-01-03
Bump cheshire to 5.6...

lsnape:
This forces the jackson.databind transitive dependency in the boot/worker project to be bumped up to 2.6, in order to fix compatability issues with s3-wagon-private version 1.3. See s3-wagon-private/s3-wagon-private#45 for more information.

The problem is the s3-wagon-private library functions are invoked within the builtin `push` task, in the context of the boot worker pod. The worker pod includes an old version cheshire, which includes a version of jackson that conflicts with s3-wagon-private.

Version 1.3 of s3-wagon-private is particularly attractive because it uses the `DefaultAWSCredentialsProviderChain`. 

To mitigate any unwanted changes I've found the earliest version (5.6) of cheshire that bumps jackson to a compatible version.

### DIFF
--- a/boot/worker/project.clj
+++ b/boot/worker/project.clj
@@ -22,7 +22,7 @@
                  [net.cgrand/parsley          "0.9.3" :exclusions [org.clojure/clojure]]
                  [mvxcvi/puget                "1.0.1"]
                  [reply                       "0.3.7"]
-                 [cheshire                    "5.3.1"]
+                 [cheshire                    "5.6.0"]
                  [clj-jgit                    "0.8.0"]
                  [clj-yaml                    "0.4.0"]
                  [javazoom/jlayer             "1.0.1"]


lsnape:
Hold fire on this. I'm still having issues locally despite bumping the jackson versions...

martinklepsch:
@lsnape thanks for looking into this and sorry you ran into this issue.

Let me know if you find a solution that seems to consistently fix the issue. 
I guess keeping these dependencies up to date isn't a bad thing either way.

lsnape:
No problem @martinklepsch!

Bumping the cheshire version does in fact fix the problem; it was my approach to testing that was wrong!

To test this, I called `set-worker-pod!` with a version of 'boot/worker', containing the upgraded cheshire, before the `push` task was invoked. However, the function [add-wagon!](https://github.com/boot-clj/boot/blob/cb923ee78ada4ade6ec60782f4b3b1762325ce30/boot/core/src/boot/core.clj#L313) loads 's3-wagon-private' on the default worker pod way earlier than that, so it was still blowing up.

A simple `make install` forces the newer 'boot/worker' to be used consistently throughout the tooling, and shows 's3-wagon-private' working as expected. 🎉

martinklepsch:
Fixed https://github.com/boot-clj/boot/commit/999a608134386840c1140a32ec99d9d2a43eca33 & https://github.com/boot-clj/boot/commit/e7155631bffdd0f209f7a98d68fc6b385e63f59d

martinklepsch:
Thanks for contributing & congrats to becoming a contributor 🙂

lsnape:
Brilliant thanks!

PR #682, opened 2018-01-09
Parse complex optargs with sequential types

### LINKED ISSUE #544: Pod `*err*` doesn't get redirected to cider
I noticed that calls to `boot.util/info` from inside a pod do not print in a cider repl.  Further investigation revealed that if you `boot repl` from a command line and `cider-connect` to it, then pod output goes to the command line, and non-pod output goes to cider.  I'm hoping there's some way to redirect pod output so that it goes to the same place.

arichiardi:
This patch enables to parse options using complex optargs like `SYM:VER=PATH`
with types that are not nested but straight up sequential, like `[sym str str]`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ and `technomancy` for the above explanation.
 - Bump [pomegranate](https://github.com/cemerick/pomegranate) and [dynapath](https://github.com/tobias/dynapath) to `1.0.0`. [#612][612]
 - Digest `java.io.File` instead `String` path of jar at sift-action `:add-jar` method [#678][678]
 - Add warning about improper use of repl :eval option [#666][666]
+- Parse complex optargs with sequential types, see [wiki section][sequential-optargs] [#682][682]
 
 #### Fixed
 
@@ -42,6 +43,7 @@ and `technomancy` for the above explanation.
 - Boot does not sign jars with classifiers [#625][625]
 - Allow clojure source jar onto the classpath [#654][654]
 - Fix `boot.tmpdir/cp` [#502][502]
+- Split an argument in three [#578][578]
 
 #### Misc
 
@@ -58,6 +60,9 @@ and `technomancy` for the above explanation.
 [678]: https://github.com/boot-clj/boot/pull/678
 [679]: https://github.com/boot-clj/boot/pull/679
 [666]: https://github.com/boot-clj/boot/pull/666
+[578]: https://github.com/boot-clj/boot/issues/578
+[682]: https://github.com/boot-clj/boot/pull/682
+[sequential-optargs]: https://github.com/boot-clj/boot/wiki/Task-Options-DSL#sequential-optargs
 
 ## 2.7.2
 
--- a/boot/core/build.boot
+++ b/boot/core/build.boot
@@ -1,17 +1,29 @@
 (set-env!
- :source-paths #{"test"}
- :dependencies '[[org.clojure/tools.reader "1.0.0-alpha2"]])
+ :source-paths #{"src" "test"}
+ :dependencies '[[org.clojure/tools.reader "1.0.0-alpha2"]
+                 [metosin/boot-alt-test "0.3.2" :scope "test"]])
+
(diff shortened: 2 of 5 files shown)


boot/core/build.boot
```diff
@@ -1,17 +1,28 @@
...
 
 (import boot.App)
 
-(ns-unmap 'boot.user 'test)
-
-(deftask test []
-  (boot.util/info "Testing against version %s\n" (App/config "BOOT_VERSION"))
-  (comp (runtests)
+(deftask integration-test []
+  (set-env! :source-paths #{"integration-test"})
```
arichiardi: I am always tripping on this one. It is not working, will fix it.

arichiardi:
Don't understand why the tests are failing in this very weird way:

```shell
FAIL in boot.cli-test/cli-tests (cli_test.clj:43)                               
parse-cli
[fix #578] split arguments with two different separators using vector optarg.
expected: {:parent [prj "1.3" "parent"]}
  actual: {:parent [ prj "1.3" "parent"]}
    diff: - {:parent [prj]}
          + {:parent [ prj]}
```

CHANGES.md
```diff
@@ -34,6 +34,7 @@ and `technomancy` for the above explanation.
 - Bump [pomegranate](https://github.com/cemerick/pomegranate) and [dynapath](https://github.com/tobias/dynapath) to `1.0.0`. [#612][612]
 - Digest `java.io.File` instead `String` path of jar at sift-action `:add-jar` method [#678][678]
 - Add warning about improper use of repl :eval option [#666][666]
+- Parse complex optargs with sequential types [682][682]
```
martinklepsch: Maybe add an example here what was previously possible and whats possible now.

boot/core/src/boot/cli.clj
```diff
@@ -43,7 +43,17 @@
...
+  "Return a function accepting arguments that are split according to optarg.
+
+  The function works also for optargs splitting:
+
+    ((parse-fn SYM:VER=PATH) SYM:VER=PATH) => [SYM VER PATH]
+
+  But the typical usage is:
+
+    ((parse-fn SYM:VER=PATH) \"prj:1.3=parent\") => [prj 1.3 parent]"
+  [optarg]
```
martinklepsch: Would this also work?
```
((parse-fn SYM:VER=PATH=) \"prj:1.3=parent=xxx\")
```

arichiardi: Probably no because the parsing needs `A-Z` interleaved by symbols at the moment

CHANGES.md
arichiardi: Ok I will add one

boot/core/src/boot/cli.clj
arichiardi: Yes it works, I tried it by having a test, one second I am going to push it

CHANGES.md
```diff
@@ -34,6 +34,7 @@ and `technomancy` for the above explanation.
 - Bump [pomegranate](https://github.com/cemerick/pomegranate) and [dynapath](https://github.com/tobias/dynapath) to `1.0.0`. [#612][612]
 - Digest `java.io.File` instead `String` path of jar at sift-action `:add-jar` method [#678][678]
 - Add warning about improper use of repl :eval option [#666][666]
+- Parse complex optargs with sequential types, see [wiki section][sequential-optargs] [#682][682]
```
martinklepsch: That's perfect 👌

martinklepsch:
Merged https://github.com/boot-clj/boot/commit/197f991324d9207f5d4b7b1828dd54ff61aadd09 https://github.com/boot-clj/boot/commit/69e2957c05023b312296f4f8940ed88b78ebaa59 https://github.com/boot-clj/boot/commit/9b21dd98d18c96d856250c0b7555b8a660f7704e

martinklepsch:
I need to revisit my workflow for merging PRs bit annoyed that this is picked up as closed instead of merged now. 😏

Issue #683, opened 2018-01-16
`println` output during lazy-seq realization in `with-eval-in` corrupts return value


martinklepsch:
```clojure
#!/usr/bin/env boot

(set-env! :dependencies '[[org.slf4j/slf4j-nop "1.7.25"]])

(require 'boot.pod)

(defn expected-result? [data]
  (println "Testing with data" data "\n")
  (let [expected   (map #(do (println "Found a match") %) data)
        pod-result (boot.pod/with-eval-in (boot.pod/make-pod {})
                     (map #(do (println "Found a match") %) ~data))]
    (when-not (= expected pod-result)
      (println "EXPECTED:   " expected)
      (println "POD RESULT: " pod-result))
    (= expected pod-result)))

(if (expected-result? [{:name "match-beginning"}])
  (do (println "Success, values are equal") (System/exit 0))
  (do (println "Failure, values are different") (System/exit 1)))
```

To reproduce: copy the above into a file `repro`, make it executable `chmod +x repro`, run it `./repro`.

I reproduced this issue with `2.6.0`, `2.7.2` and `2.8.0-SNAPSHOT`.

**EDIT** updated with more minimal repro. The key thing seems to be that something is printed during iteration over a sequence.

martinklepsch:
FWIW this seems to fix the issue, not sure why yet though (note the double `pr-str`)
```diff
diff --git a/boot/pod/src/boot/pod.clj b/boot/pod/src/boot/pod.clj
index b4f58f6..6903737 100644
--- a/boot/pod/src/boot/pod.clj
+++ b/boot/pod/src/boot/pod.clj
@@ -465,7 +473,9 @@
   ([expr]
      (let [{:keys [meta? expr]} (read-string expr)]
        (binding [*print-meta* meta?]
-         (pr-str (eval expr)))))
+         (let [r (eval expr)] 
+           (pr-str r)
+           (pr-str r)))))
   ([pod expr]
      (let [arg (pr-str {:meta? *print-meta* :expr expr})
            ret (with-invoke-in pod (boot.pod/eval-in* arg))]
```

martinklepsch:
Laziness seems to be at fault here. Using `mapv` in the example below solves the issue but this doesn't really help when lazy seqs are nested deep inside maps etc. 
```clojure
(->> (range 30)
     (map #(do (println %)))
     (pr-str))
```

martinklepsch:
```clojure
(->> (range 30)
     (map #(do (println %) %))
[...]

PR #684, opened 2018-01-16
Fully serialize pod return values before pr-str'ing

martinklepsch:
Fix for #683.

I added a test as well but `lein test` isn't working. Will need to look into that.

### DIFF
--- a/boot/pod/src/boot/pod.clj
+++ b/boot/pod/src/boot/pod.clj
@@ -2,6 +2,7 @@
   (:require
     [clojure.set                  :as set]
     [clojure.string               :as string]
+    [clojure.walk                 :as walk]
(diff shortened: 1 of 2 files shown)


boot/pod/src/boot/pod.clj
```diff
@@ -465,7 +466,9 @@
   ([expr]
      (let [{:keys [meta? expr]} (read-string expr)]
        (binding [*print-meta* meta?]
-         (pr-str (eval expr)))))
+         (->> (eval expr)
+              (walk/postwalk identity) ; make sure all lazy seqs are realized #683
```
arichiardi: LGTM

arichiardi:
If we merge #685 in then this can use it :smile:

PR #685, opened 2018-01-16
Introduce build.boot and boot-alt-test in boot/pod

arichiardi:
This complete the mini migration of our tests to boot. Note that clojure 1.7.0
has to be used in build.boot or boot-alt-test won't run (to be investigated).


martinklepsch:
Thanks! 🙌  I considered asking you for this but felt like I shouldn't 🙂

arichiardi:
Not working for now...something weird going on. I though it was an easy one but no ... :smile:

PR #696, opened 2018-05-25
Add JDK 9 and cache dependencies in Travis

tirkarthi:
* Add JDK 9 to Travis for testing
* Cache dependencies in Travis which will reduce the load on Maven and Clojars since the build downloads around 300 artifacts

Successful build : https://travis-ci.org/tirkarthi/boot/builds/383569211


alandipert:
Thank you

Issue #698, opened 2018-06-05
Build failures in JDK 11


tirkarthi:
### Problem Description

Tests fails in Oracle JDK 11

### Steps to reproduce

* Run tests agains JDK 11 in travis by adding `oraclejdk11` to `.travis.yml`
* Build failure : https://travis-ci.org/tirkarthi/boot/jobs/388238099

This is due `java.xml.bind` which was removed in JDK 9 . I don't know why it builds in JDK 9 and 10 but throws exception in JDK 11.

Relevant source code : https://github.com/boot-clj/boot/blob/30ba3ce69dcd8b2b3e00677a14db05c8ce49d59c/boot/base/src/main/java/boot/App.java#L124

tirkarthi:
Ref : http://openjdk.java.net/jeps/320

In Java SE 9, the Java SE modules that contain Java EE and CORBA technologies are annotated as deprecated for removal, indicating the intent to remove them in a future release:

    java.xml.ws (JAX-WS, plus the related technologies SAAJ and Web Services Metadata)
    java.xml.bind (JAXB)
    java.activation (JAF)
    java.xml.ws.annotation (Common Annotations)
    java.corba (CORBA)
    java.transaction (JTA)

Related modules in Java SE 9 are also deprecated for removal:

    java.se.ee (Aggregator module for the six modules above)
    jdk.xml.ws (Tools for JAX-WS)
    jdk.xml.bind (Tools for JAXB)

Since deprecating modules for removal merely causes compile-time warnings, JDK 9 took a more robust step to prepare developers for the actual removal of these modules in a future release: The modules are not resolved in JDK 9 when code on the class path is compiled or run. This allows developers on JDK 9 to deploy standalone versions of the Java EE and CORBA technologies on the class path, just like on JDK 8. Alternatively, developers on JDK 9 can use the --add-modules flag on the command line to resolve the modules in the JDK runtime image.

PR #699, opened 2018-06-05
Test against JDK 9 and 10

tirkarthi:
Travis added support for JDK 9, 10 and 11 along with EA versions. I have added JDK 9 and 10 for testing. I have removed JDK 11 since there are build failures. I have also added cache for `~/.m2` which reduces load on Maven and Clojars.

I have raised a separate issue for JDK 11 at #698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
+  - oraclejdk10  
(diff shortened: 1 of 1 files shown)


tirkarthi:
Sorry, I just realized I already made a PR with #696 . I will add JDK 10 as part of that. Closing this one. Sorry for the noise and mistake from my part.

PR #703, opened 2018-06-26
Update nREPL to 0.4.4

### LINKED ISSUE #695: Migration plans for nREPL 0.4
That's basically a follow up of nrepl/nREPL#1 and was also discussed in nrepl/nREPL#4.

TL;DR; - In verions 0.4 of the "new" nREPL the namespaces will be changed from `clojure.tools.nrepl.*` to `nrepl.*`. 

As this is a different artifact technically speaking(`nrepl/nrepl` instead of `clojure.tools/nrepl`) I think the best migration path for Boot would be to just start depending on both projects and have two tasks for the two nREPLs until we can drop the old version (e.g. `boot repl` and `boot nrepl`). The second task can just start a headless server until reply adds support for `nrepl/nrepl`. 

How does this sound?

bbatsov:
As no one really commented on #695  I've decided to try an alternative approach to get some conversation going. Here's my preferred (quick) solution to the migration to the new nREPL. I know this is probably less than ideal, but it's a one time deal, and I'd really appreciate your support as I rather be dealing with nREPL improvements than with migration paths.

nREPL was recently moved out of clojure-contrib and its development
continued under a dedicated org (https://github.com/nrepl). The new project is (and will remain
to be) completely backwards compatible on the nREPL API level, but
the namespace had to be changed.

This fixes #695.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,16 @@
 
 [702]: https://github.com/boot-clj/boot/pull/702
 
+#### Breaking
+
+Updated nREPL to 0.4.4. The new nREPL has a new artefact coordinates
+(`nrepl/nrepl`) and a new namespace prefix (`clojure.tools.nrepl.` ->
+`nrepl.`). While the nREPL protocol is 100% compatible with that of
+the 0.2.x series you'll have to make sure that any middleware you're
+using has been updated to target 0.4.x. You can find more details
+[here](https://github.com/nrepl/nREPL/issues/1) and
+[here](https://github.com/nrepl/nREPL/blob/master/HISTORY.md).
+
 ## 2.8.1
 
 #### Improved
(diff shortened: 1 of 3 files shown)


alandipert:
Oh, this could use a CHANGELOG.md entry though. Thanks in advance for adding one that summarizes the change.

arichiardi:
Just for future reference, this is not really backward compatible, see:
https://github.com/nrepl/nREPL#status

However the tooling goal would be to move towards it we could bump `boot`'s major and advertise the change way in advance. In the meantime folks (like myself) can try things out with all the `master` code bases (many involved) and confirm everything works.

bbatsov:
Changelog entry added.

bbatsov:
PR updated for nREPL 0.4.3.

CHANGES.md
```diff
@@ -8,6 +8,16 @@
 
 [702]: https://github.com/boot-clj/boot/pull/702
 
+#### Breaking
+
+Updated nREPL to 0.4.3. The new nREPL has a new artefact coordinates
```
thenonameguy: `s/artefact/artifact` ?

bbatsov: Not in my part of the world. 😆

thenonameguy: Sure, just noted it to keep it in line with maven spelling:
screen shot 2018-07-26 at 4 21 09 pm

Also: https://github.com/boot-clj/boot/search?q=artifact&unscoped_q=artifact
vs.
https://github.com/boot-clj/boot/search?q=artefact&unscoped_q=artefact

bbatsov: Yeah, yeah. I know everyone in programming uses American English spellings, but it's hard for me to remember this all the time. Especially when my spell checker highlights `artifact` as incorrect. :-)

bbatsov:
The build failure is unrelated, btw.

@alandipert Can you tell me if you want us to move forward with this or not? I've invested a tremendous amount of time already coordinating the nREPL upgrade across all related projects and that's the final missing piece. I'd be extremely happy if we can just wrap this up and I can focus on actually improving the server and its middleware. :-)

bbatsov:
Thanks!

PR #724, opened 2019-01-01
Update nREPL and REPL-y

bbatsov:
A straightforward change.

One thing to note - nREPL introduced a config file, that ideally boot should support at some point. See https://nrepl.org/nrepl/usage/server.html#_server_options

It's not a big deal and should be trivial to do. Perhaps someone like @alexander-yakushev would like to land a hand with this, as I'm not really a boot user.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## 2.8.3
+
+#### Improved
+
+- Updated REPL-y to 0.4.3.
+
 ## 2.8.2
 
 #### Fixed
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 #### Improved
 
 - Updated REPL-y to 0.4.3.
+- Updated nREPL to 0.5.3.
 
 ## 2.8.2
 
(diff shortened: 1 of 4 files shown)


boot/pod/src/boot/repl.clj
```diff
@@ -8,7 +8,7 @@
     [boot.from.io.aviso.exception :refer [*fonts*]]))
 
 (def ^:dynamic *default-dependencies*
-  (atom '[[nrepl/nrepl "0.4.4" :exclusions [[org.clojure/clojure]]]]))
+  (atom '[[nrepl/nrepl "0.5.3"]]))
```
thenonameguy:
```suggestion
  (atom '[[nrepl/nrepl "0.6.0"]]))
```

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # Changes
 
+## 2.8.3
+
+#### Improved
+
+- Updated REPL-y to 0.4.3.
+- Updated nREPL to 0.5.3.
```
thenonameguy:
```suggestion
- Updated nREPL to 0.6.0.
```

bbatsov:
@flyboarder Can you merge this and bump the version to 0.6.0? Tools like CIDER now target nREPL 0.6 and it’d be nice if Boot users didn’t have to manually override their nREPL dependency.

flyboarder:
Yes, I’ll look into a new patch version soon.

bbatsov:
Thanks! 

Generally nREPL upgrades are safe, as we’re very careful about backwards compatibility.

PR #737, opened 2019-01-27
Fix Travis

flyboarder:


bbatsov:
Btw, it might be more prudent to migrate to Circle CI in light of the uncertain future of Travis (new sketchy owner, layoffs, etc). I've been migrating all my projects to Circle recently and I'm really pleased with the results.

PR #746, opened 2019-05-13
Use filesystem events directly in watcher

jeroenvandijk:
The current file watcher is very slow on large projects.  This change is a naive approach to use the available filesystem events directly in order to prevent the costly rescanning of the filesystem.

## Motivation and Context
The file watcher is a fundamental part of the virtual filesystem that Boot implements. When there is a delay in updating this filesystem this delay is forwarded to any system that depends on responding to handling file updates. Especially when an external system is used for file updates this becomes visible (e.g. [Hawk](https://github.com/wkf/hawk)). For small projects this is somewhat manageable with correctly timed `Thread/sleep`s [after being bitten by random failures due to the subtle delay], for larger projects these updates become too slow and the delay is noticable and annoying.

## How Has This Been Tested?
I've tested it manually with an autotest script to see if the latency was improved. It is unclear if there are edge cases that are now broken.

My environment is Mac OSX (High Sierra). Since this change might behave differently accross filesystems it needs to be tested in other environments too.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/boot/core/src/boot/core.clj
+++ b/boot/core/src/boot/core.clj
@@ -726,6 +726,14 @@
 
 ;; Boot Environment ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(require '[clojure.edn])
+
+(defn- take-changes [debounce q]
+  (loop [init {}]
+    (if-let [more (.poll q debounce TimeUnit/MILLISECONDS)]
+      (recur (merge-with clojure.set/union init (clojure.edn/read-string more)))
+      init)))
+
 (defn watch-dirs
   "Watches dirs for changes and calls callback with set of changed files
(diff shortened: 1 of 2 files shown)


jeroenvandijk:
Hmm i think my "fix" doesn't add a (noticable) improvement. I just found out it's actually my editor (emacs) that is slow in saving my files :see_no_evil: With Textmate for both the old code and new code the effect is instant, with emacs there is a delay in both cases.

